PR #20, opened 2023-10-10
ci: use release-please

Billlynch:

Use release please to create PRs for our releases, and github releases.

- [ ] All tests are passing
- [ ] linter/sytle run on changed files


.github/workflows/Build.yml
```diff
@@ -0,0 +1,45 @@
+name: Build, Test, List
```
nicklasl, line 1:
```suggestion
name: Build, Test, Lint
```

.github/workflows/Release-Please.yml
```diff
@@ -0,0 +1,67 @@
...
+
+  Setup-Environment:
+    runs-on: ubuntu-latest
+    if: ${{ steps.release.outputs.releases_created }}
+    strategy:
+      matrix:
+        node-version: [ 18.x ]
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
```
nicklasl, line 36: Could be a good idea to pass the `ref` to access the tag that was created by release please.  
```
with:
  ref: ${{ needs.release-please.outputs.release_tag_name }}
// or 
   ref: ${{ steps.release.outputs.release_tag_name }}
```

lerna.json
```diff
@@ -3,6 +3,7 @@
   "version": "independent",
   "packages": ["packages/*"],
   "npmClient": "yarn",
+  "tagVersionPrefix": "-v",
   "command": {
     "version": {
```
andreas-karlsson, line 8: This is not needed anymore right? Cause now release-please is doing changelog and version management?

.github/workflows/Release-Please.yml
```diff
@@ -0,0 +1,67 @@
...
+        run: yarn install --frozen-lockfile
+
+  Build:
+    runs-on: ubuntu-latest
+    if: ${{ steps.release.outputs.releases_created }}
+    needs: [ Setup-Environment ]
+    steps:
+      - run: yarn build
+
+  npm-publish:
```
andreas-karlsson, line 51: How does this work? Configured in this way, is npm-publish dependent on Build? Or should build be a step in Publish? I guess the same would be true for the release-please job, so maybe I'm misunderstanding actions..

Billlynch: The 'needs' bit of the configuration means that it will wait until that job has completed

lerna.json
Billlynch: yes, but lerna is still doing the publish. release please uses <component-name>-v<semver> wheras lerna uses <component-name>@<semver> by default. As you can't change the tags as much in release please (only to use the v or not and the component name) i've changed the lerna tags to match the release please tags

Billlynch: oh i see, version may not be needed, let me check

.github/workflows/Commitlint.yml
```diff
@@ -5,7 +5,7 @@ on:
     branches: [main]
 
 jobs:
-  build:
+  check-commit-message:
```
andreas-karlsson, line 8: nit, but would be nice to stick with a consistent naming convention, so:
```suggestion
  Check-Commit-Message:
```

andreas-karlsson (approved):
Nice! Just a minor comment on job naming convention if you will.

andreas-karlsson (dismissed):

andreas-karlsson (approved):

PR #27, opened 2023-10-12
refactor(client-http,web,server): use a serializable and more simple …

Billlynch:
…configuration object

Refactor to use a more simple configuration object. The aim to be simple to integrate with any adaptor when using the client-http directly. A key aim was also to ensure that the object is serialisable to enable better Next.js integrations down the road.

- [x] All tests are passing
- [x] linter/sytle run on changed files


packages/client-http/src/client/Configuration.ts
```diff
@@ -8,18 +8,72 @@ export namespace Configuration {
...
+    value: T;
+    schema: FlagSchema;
+  }
+
+  export namespace FlagValue {
+    export function matches<T>({ schema }: FlagValue<T>, value: any): value is T {
+      return valueMatchesSchema(value, schema);
+    }
+
+    export type Traversed<T, S extends string> = S extends `${infer STEP}.${infer REST}`
```
andreas-karlsson, line 37: I think I found a problem in this type. If it's used on `FlagValue<unknown>` I'd expect the result to always be `unknown` but now it is `never`.

packages/client-http/src/client/Configuration.ts
```diff
@@ -8,18 +8,72 @@ export namespace Configuration {
...
+  export interface Flag<T = unknown> extends FlagValue<T> {
+    name: string;
+    reason: ResolveReason;
+    variant: string;
+    value: T;
+    schema: FlagSchema;
+  }
+
+  export namespace FlagValue {
+    export function matches<T>({ schema }: FlagValue<T>, value: any): value is T {
```
andreas-karlsson, line 33: Should this be the other way around? We use the type of value to infer T? I have an example of that locally if you want to sync.

packages/openfeature-server-provider/src/ConfidenceServerProvider.ts
```diff
@@ -67,40 +68,34 @@ export class ConfidenceServerProvider implements Provider {
...
-          reason: flag.reason,
+          reason: 'ERROR',
         };
       }
+
       this.applyManager.apply(configuration.resolveToken, flagName);
       return {
-        value: flagValue.value,
-        reason: 'TARGETING_MATCH',
+        value: flagValue.value as T,
```
andreas-karlsson, line 97: This cast can be removed if we update the match function according to above.

packages/client-http/src/client/ConfidenceClient.ts
```diff
@@ -17,7 +16,20 @@ type ResolveResponse = {
...
+  schema: {
+    [key: string]: ConfidenceBaseTypes | { structSchema: ConfidenceFlagSchema };
+  };
+};
 
+export type ResolvedFlag<T = any> = {
+  flag: string;
+  variant: string;
+  value?: T;
+  flagSchema?: ConfidenceFlagSchema;
```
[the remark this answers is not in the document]
Billlynch: if you resolve a flag without any context then the response will not have a value or a schema. But will have "" for variant and a no segment match reason. So the reason is because it matches the possible return values from the backend.

andreas-karlsson (dismissed):

andreas-karlsson (approved):

PR #32, opened 2023-10-25
ci(examples): on release please version update bump example deps

Billlynch:

Added a git hook to bump the version number of the mono-repo packages in the example apps.
Added a checkmark to the PR template to remind people to use the example apps to verify changes.

When developing you will be using the current version number but the locally running example app will be using code built from the source in the mono-repo. Hence, when release please creates an update we are safe (assuming the changes have been tested in the example app, as they should be) to bump the version numbers in the example apps too using a hook.

- [ ] All tests are passing


nicklasl (approved):

scripts/bump-example-deps.js
```diff
@@ -0,0 +1,39 @@
...
+      continue;
+    }
+
+    if (currentPackageVersion === releasePleaseManifestJSON[dep]) {
+      console.log('No update needed for: @spotify-confidence/%s in %s', depName, packageJson.name);
+      continue;
+    }
+
+    packageJson.dependencies[`@spotify-confidence/${depName}`] = releasePleaseManifestJSON[dep];
+    fs.writeFileSync(packageJsonPath, `${JSON.stringify(packageJson, null, 2)}\n`);
```
andreas-karlsson, line 29: NIT: would feel nicer to just write once after all deps are updated.

andreas-karlsson (approved):
Nice. Just a small comment, but it that was intentional I'm fine with how it is.

scripts/bump-example-deps.js
Billlynch: agree this can be improved, we can add a ticket to improve this later i think! :)

PR #33, opened 2023-10-25
Apply improvements

Billlynch:

Adding some docs around the concept of apply.

WEB:
Set access apply to the default for the web provider.
Add docs around the configuration of apply.

SERVER
Remove ability to use access apply as it doesn't make sense / would have caused issues in a server context. 

- [x] All tests are passing


packages/openfeature-server-provider/src/factory.ts
```diff
@@ -9,18 +9,13 @@ type ConfidenceProviderFactoryOptions = {
...
   baseUrl?: string;
-  apply?: {
-    timeout: number;
-  };
 };
 
 export function createConfidenceServerProvider(options: ConfidenceProviderFactoryOptions): Provider {
   const confidenceClient = new ConfidenceClient({
     ...options,
-    apply: !options.apply,
```
andreas-karlsson, line 20: Could it be the case that users don't want apply at all? For instance if they do their own analysis and don't want to pay for the extra resources used by apply? No need to include in this PR though.

packages/openfeature-web-provider/src/ConfidenceWebProvider.ts
```diff
@@ -15,9 +15,11 @@ import equal from 'fast-deep-equal';
 import { ApplyManager, ConfidenceClient, Configuration, ResolveContext } from '@spotify-confidence/client-http';
 
 export interface ConfidenceWebProviderOptions {
-  apply?: {
-    timeout?: number;
-  };
+  apply:
+    | {
+        timeout: number;
```
andreas-karlsson, line 20: Should we actually have min-value for timeout? Can generate a lot of requests and it will be difficult to enforce later. I actually think 250ms is a good minimum. Maybe we should even have it non configurable for now. I.e. just `backend` or `client`

packages/openfeature-server-provider/src/factory.ts
Billlynch: interesting. potentially yes. Maybe something we can tackle in a future PR :)

packages/openfeature-web-provider/src/ConfidenceWebProvider.ts
Billlynch: After a sync, we decided to keep it simple for now and use client and backend. where client has a timeout of 250.

Billlynch:
I'll squash the fixup(s) when the PR is approved :)

packages/openfeature-web-provider/README.md
```diff
@@ -46,3 +46,20 @@ Notes:
 
 - It's advised not to perform `setContext` while `setProvider` is running, you can await setting the context first, or listen to the `ProviderEvent.Ready` via a handler on `OpenFeaure`.
 - It's advised not to perform resolves while `setProvider` and `setContext` are running: resolves might return the default value with reason `STALE` during such operations.
+
+## Configuring Apply
+
+See [apply concept](../../concepts/apply.md).
+
+By default, access apply is used, using a timeout of 250ms. However, this can be configured by supplying a timeout
```
andreas-karlsson, line 54: Setting a timeout is not true anymore right?

concepts/apply.md
```diff
@@ -0,0 +1,48 @@
...
+These are subtly different, depending on the use-case one is preferred over the other.
+
+## Backend Apply
+
+Applying on flag request introduces no additional work on the application. When the flag is requested from the
+Confidence Server it is marked as applied. This is regardless of if the value of the flag was used in the code.
+
+Backend apply is recommended for server applications, as we also recommend requesting a flag every time a flag is used.
+
+## Access Apply
```
andreas-karlsson, line 27: Should we just call it `Client Apply`? Or we could have the options being `backend | access`? I'm fine with both, but mixing the wording seems a bit confusing.

Billlynch: i think access, but good shout

andreas-karlsson (approved):

PR #37, opened 2023-11-09
Max buffer size

### LINKED ISSUE #36: Max Buffer Size for Apply Events
If many flags are requested within 250ms of each other when using 'client' apply logic the buffer has no max size. This will result in the apply event not being sent and an unbounded buffer may cause out of memory errors in the application. 

The buffer of apply event should have a max size.

A max buffer size should be set (in terms of flags) for simplicity. This would cancel the timeout, flush the buffer and send the event when hit.

## Steps to Reproduce

1. Step 1
2. Step 2
3. ...

## Your Environment

- NodeJS Version (v18):
- Operating System and Version (e.g. Ubuntu 14.04):
- Browser Information:

Billlynch:

closes #36 by setting a max size of the apply buffer, only applicable in web.

Need to update the docs when the apply documentation PR is merged. (#33 )

- [x] All tests are passing

### DIFF
--- a/packages/client-http/src/ApplyManager.ts
+++ b/packages/client-http/src/ApplyManager.ts
@@ -2,6 +2,7 @@ import { ConfidenceClient, AppliedFlag } from './client';
 
 export interface ApplyManagerOptions {
   timeout: number;
+  maxBufferSize: number;
   client: ConfidenceClient;
 }
 
@@ -9,11 +10,13 @@ export class ApplyManager {
   private resolveTokenPending: Map<string, AppliedFlag[]> = new Map();
   private resolveTokenSeen: Map<string, Set<string>> = new Map();
   private readonly timeout: number;
+  private readonly maxBufferSize: number;
   private client: ConfidenceClient;
   private flushTimeout: ReturnType<typeof setTimeout> | null = null;
 
   constructor(options: ApplyManagerOptions) {
     this.timeout = options.timeout;
+    this.maxBufferSize = options.maxBufferSize;
     this.client = options.client;
   }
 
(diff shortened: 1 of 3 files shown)


packages/openfeature-web-provider/src/ConfidenceWebProvider.ts
```diff
@@ -16,6 +16,7 @@ import equal from 'fast-deep-equal';
 import { ApplyManager, ConfidenceClient, Configuration, ResolveContext } from '@spotify-confidence/client-http';
 
 const APPLY_TIMEOUT = 250;
+const MAX_APPLY_BUFFER_SIZE = 20;
```
Billlynch, line 19: This needs to be discussed, I think this is an ok default. 

We should allow this to be set in the factory though

fabriziodemaria: This is the same batch size we chose for our mobile Providers. I think it's ok as default. Not sure it's worth to make this configurable by the users, as `apply` events are managed by us end-to-end

packages/client-http/src/ApplyManager.ts
```diff
@@ -61,6 +64,11 @@ export class ApplyManager {
     if (this.flushTimeout) {
       clearTimeout(this.flushTimeout);
     }
-    this.flushTimeout = setTimeout(() => this.flush(), this.timeout);
+
+    if ((this.resolveTokenPending.get(resolveToken)?.length || 0) >= this.maxBufferSize) {
+      this.flush();
+    } else {
+      this.flushTimeout = setTimeout(() => this.flush(), this.timeout);
```
fabriziodemaria, line 71: Does this mean that we wait until the buffer is full, before sending `apply`? If so, I think it's safer to try send all the apply data more often to avoid cases where an half-full buffer will never get sent

packages/openfeature-web-provider/src/ConfidenceWebProvider.ts
Billlynch: happy to mirror the mobile sdks for this then :)

packages/client-http/src/ApplyManager.ts
Billlynch: nope! This means that if the buffer is full it will call flush (i.e. send the apply event). If the buffer is not full it will wait to see if anything else can be put in the buffer before the timeout ends, then send. In other words, the buffer size will call flush, otherwise act as before :)

fabriziodemaria (approved):

PR #39, opened 2023-11-09
Send SDK version 

Billlynch:

We can now send the sdk version in the resolve request. This PR enables this for the web and server providers

- [x] All tests are passing

### DIFF
--- a/packages/openfeature-web-provider/src/factory.ts
+++ b/packages/openfeature-web-provider/src/factory.ts
@@ -15,6 +15,10 @@ export function createConfidenceWebProvider(options: ConfidenceWebProviderFactor
   const confidenceClient = new ConfidenceClient({
     ...options,
     apply: options.apply === 'backend',
+    sdk: {
+      id: 'SDK_ID_JS_WEB_PROVIDER',
+      version: '0.0.0', // x-release-please-version
+    },
   });
 
   return new ConfidenceWebProvider(confidenceClient, {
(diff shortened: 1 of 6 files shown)


packages/openfeature-server-provider/src/version.ts
```diff
@@ -0,0 +1 @@
+export const VERSION = '0.0.5';
```
Billlynch, line 1: I'm looking at getting release-please to update this using the extra-files config. I don't want to import from the package.json because it would mean that would need to be in the bundle, and also would require some more brittle IO. I think this is safer for a library running in the browser

nicklasl: That should just be [here](https://github.com/spotify/confidence-openfeature-provider-js/blob/main/release-please-config.json#L22), right? Just not very certain if the file location is meant to be relative to the package or not.

Billlynch: I think it's package, based off how the readme is accessed, also had to add the comment line for the generic updater. I'm trialing this in my test repo! :)

Billlynch: verified this approach will work, and it is package scoped

fabriziodemaria (approved):
LGTM

andreas-karlsson (dismissed):

packages/openfeature-web-provider/src/factory.ts
```diff
@@ -14,6 +14,10 @@ export function createConfidenceWebProvider(options: ConfidenceWebProviderFactor
   const confidenceClient = new ConfidenceClient({
     ...options,
     apply: options.apply === 'backend',
+    sdk: {
+      id: 'SDK_ID_JS_WEB_PROVIDER',
+      version: '0.0.0', // x-release-please-version
```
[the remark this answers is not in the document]
Billlynch: yes! tested in my actions-test repo :)

andreas-karlsson (approved):

PR #47, opened 2023-11-21
refactor get flag to throw open feature errors and handle targeting key error

Billlynch:
Currently we set default value, error code and error message manually in the providers, this can make reading the getFlag function more difficult. OpenFeature's SDK will catch OF errors and map these in the exact same way. So we can simplify the logic with these errors. This is pre-work to make apply logic easier to adjust in the near future. 

Also added throwing invalid context error when confidence returns with a reason of targeting key error

- [x] All tests are passing

### DIFF
--- a/packages/openfeature-server-provider/src/ConfidenceServerProvider.test.ts
+++ b/packages/openfeature-server-provider/src/ConfidenceServerProvider.test.ts
@@ -1,4 +1,11 @@
-import { FlagNotFoundError, Logger, ParseError, ProviderStatus, TypeMismatchError } from '@openfeature/js-sdk';
+import {
+  FlagNotFoundError,
+  Logger,
+  ParseError,
+  ProviderStatus,
+  TypeMismatchError,
+  InvalidContextError,
+} from '@openfeature/js-sdk';
 import { ConfidenceClient, Configuration } from '@spotify-confidence/client-http';
 import { ConfidenceServerProvider } from './ConfidenceServerProvider';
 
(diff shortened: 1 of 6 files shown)


packages/openfeature-server-provider/src/ConfidenceServerProvider.test.ts
```diff
@@ -65,6 +72,13 @@ const dummyConfiguration: Configuration = {
       value: undefined,
       schema: 'undefined',
     },
+    ['targeting-error-flag']: {
+      name: 'targeting-error-flag',
```
fabriziodemaria, line 76: Perhaps `targeting-key-error-flag`?

packages/openfeature-web-provider/src/ConfidenceWebProvider.test.ts
```diff
@@ -267,67 +296,34 @@ describe('ConfidenceProvider', () => {
...
-
-      expect(actual).toEqual({
-        value: false,
-        errorCode: ErrorCode.PROVIDER_NOT_READY,
-        reason: 'ERROR',
-      });
+    it('should throw if the provider is not ready', () => {
+      expect(() =>
+        instanceUnderTest.resolveBooleanEvaluation('testFlag.bool', false, dummyEvaluationContext, dummyConsole),
+      ).toThrow(new GeneralError('Provider not ready'));
```
fabriziodemaria, line 302: Seems like `PROVIDER_NOT_READY` [error](https://openfeature.dev/specification/types#error-code) has not been added [to the SDK](https://github.com/open-feature/js-sdk/tree/main/packages/shared/src/errors) yet. Perhaps we can open a PR in the SDK repo and then throw the related exception here?

fabriziodemaria (approved):
Approving, although to be more compliant to the OF specification we should probably address my comment?

packages/openfeature-web-provider/src/ConfidenceWebProvider.test.ts
Billlynch: Yes, i have that change ready to go actually! :)

PR #49, opened 2023-11-22
ci: run build on all lts node versions supported

Billlynch:

Making sure we build and test on all node lts versions above our min supported. (which is dictated by open feature)  

- [ ] All tests are passing

### DIFF
--- a/.github/workflows/Build.yml
+++ b/.github/workflows/Build.yml
@@ -10,7 +10,7 @@ jobs:
   Build-And-Test:
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


Billlynch:
we have a min node version of 16.x (matching OpenFeature) however one of our other deps (spotify eslint ) has a min version of 18.x. 

16.x is no longer getting support, so i think we could realistically up our min version to 18.x. Or we could use an older version of spotify eslint

andreas-karlsson:
Won't do this for now.

PR #53, opened 2024-01-19
fix: Duplicate apply bug

andreas-karlsson:
## Fix duplicate apply on buffer overflow
Before apply events could be duplicated if the ApplyManager was flushed (which happens on buffer overflow) while a apply request was already in progress. This is fixed by clearing pendingApply while a request is in progress and adding them back if the request fails. Also refactored the code a bit to make it easier to follow.

The apply logic still has some shortcomings, but these will be adressed in a followup PR.

- [x] All tests are passing

### DIFF
--- a/packages/client-http/src/ApplyManager.ts
+++ b/packages/client-http/src/ApplyManager.ts
@@ -7,8 +7,8 @@ export interface ApplyManagerOptions {
 }
 
 type TokenRecord = {
-  pendingApply: AppliedFlag[];
-  seenFlags: Set<string>;
+  readonly pendingApply: AppliedFlag[];
+  readonly seenFlags: Set<string>;
 };
 
 export class ApplyManager {
@@ -26,6 +26,7 @@ export class ApplyManager {
 
   async flush(): Promise<void> {
     for (const [resolve_token, { pendingApply }] of Array.from(this.tokenRecords.entries())) {
+      if (!pendingApply.length) continue;
       // remove all flags while trying
       const flagsToSend = pendingApply.splice(0, pendingApply.length);
       this.client.apply(flagsToSend, resolve_token).catch(() => {
--- a/packages/client-http/src/ApplyManager.ts
+++ b/packages/client-http/src/ApplyManager.ts
@@ -6,9 +6,13 @@ export interface ApplyManagerOptions {
   client: ConfidenceClient;
 }
 
+type TokenRecord = {
+  pendingApply: AppliedFlag[];
+  seenFlags: Set<string>;
+};
+
 export class ApplyManager {
-  private resolveTokenPending: Map<string, AppliedFlag[]> = new Map();
-  private resolveTokenSeen: Map<string, Set<string>> = new Map();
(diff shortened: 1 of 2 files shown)


packages/client-http/src/ApplyManager.ts
```diff
@@ -21,54 +25,46 @@ export class ApplyManager {
...
-
-      if (flagsToSend.length === 0) {
-        continue;
-      }
-
-      try {
-        this.client.apply(flagsToSend, resolve_token).then(() => {
-          this.resolveTokenPending.set(resolve_token, []);
-        });
-      } catch (e: unknown) {
```
andreas-karlsson, line 35: client.apply is an async method so it cannot throw.

packages/client-http/src/ApplyManager.ts
```diff
@@ -6,9 +6,13 @@ export interface ApplyManagerOptions {
...
 
+type TokenRecord = {
+  pendingApply: AppliedFlag[];
+  seenFlags: Set<string>;
+};
+
 export class ApplyManager {
-  private resolveTokenPending: Map<string, AppliedFlag[]> = new Map();
-  private resolveTokenSeen: Map<string, Set<string>> = new Map();
+  private readonly tokenRecords: Map<string, TokenRecord> = new Map();
```
andreas-karlsson, line 15: Since both these maps were keyed by resolveToken I baked them into one using the new TokenRecord type.

packages/client-http/src/ApplyManager.ts
```diff
@@ -21,54 +25,46 @@ export class ApplyManager {
...
-        this.client.apply(flagsToSend, resolve_token).then(() => {
-          this.resolveTokenPending.set(resolve_token, []);
-        });
-      } catch (e: unknown) {
-        // TODO log errors
-      }
+    for (const [resolve_token, { pendingApply }] of Array.from(this.tokenRecords.entries())) {
+      // remove all flags while trying
+      const flagsToSend = pendingApply.splice(0, pendingApply.length);
+      this.client.apply(flagsToSend, resolve_token).catch(() => {
```
nicklasl, line 31: let's exit early for `[]`.

packages/client-http/src/ApplyManager.ts
```diff
@@ -6,9 +6,13 @@ export interface ApplyManagerOptions {
   client: ConfidenceClient;
 }
 
+type TokenRecord = {
+  pendingApply: AppliedFlag[];
+  seenFlags: Set<string>;
```
nicklasl, line 11: maybe make these readonly.

packages/client-http/src/ApplyManager.ts
nicklasl, line 31: Let's add a `then` block where we can push flags back for non 200 responses.

nicklasl: not for now

nicklasl (approved):

Billlynch (dismissed):

nicklasl (approved):

PR #69, opened 2024-02-26
Use a default fetch implementation

andreas-karlsson:

This PR updates all tests to use the global fetch implementation, both in browsers and NodeJS. Before we were using a mix of different hacked fetch implementations.

It also makes the `fetchImplementation` setting optional when creating a provider, and defaults this setting to use the global fetch, and throwing an error if it doesn't exist.

- [ ] All tests are passing

### DIFF
--- a/packages/openfeature-server-provider/src/ConfidenceServerProvider.e2e.test.ts
+++ b/packages/openfeature-server-provider/src/ConfidenceServerProvider.e2e.test.ts
@@ -1,36 +1,18 @@
-import { OpenFeature, ProviderEvents } from '@openfeature/server-sdk';
-import axios from 'axios';
+import { OpenFeature } from '@openfeature/server-sdk';
 import { createConfidenceServerProvider } from './factory';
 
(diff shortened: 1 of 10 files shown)


packages/openfeature-server-provider/src/ConfidenceServerProvider.e2e.test.ts
```diff
@@ -1,36 +1,18 @@
...
-        return await axios.post(url as string, request?.body).then(
-          resp =>
-            ({
-              json: async () => {
-                return resp.data;
-              },
-            } as Response),
-        );
-      },
+      fetchImplementation: fetch,
```
vahidlazio: what's changing that previously we were waiting for the ready signal and now we are not?

andreas-karlsson: I don't know why we were waiting for the ready signal before. This is dynamic context (server) so it doesn't really make sense to wait for ready.

PR #73, opened 2024-03-13
Total confidence: wip

nickybondarenko:

- [x ] All tests are passing
- [x ] linter/style run on changed files
- [x ] Tests added for new functionality

### DIFF
--- a/packages/sdk/src/Confidence.test.ts
+++ b/packages/sdk/src/Confidence.test.ts
@@ -20,4 +20,50 @@ describe('Confidence', () => {
       expect(confidence.getContext()).toEqual({ clothes: { pants: 'yellow' } });
     });
   });
+  describe('setContext', () => {
+    it('sets context', () => {
+      const confidence = new Confidence({} as any);
+      const newContext = {
+        pants: 'yellow',
+      };
+      confidence.setContext(newContext);
+      expect(confidence.getContext()).toEqual(newContext);
+    });
+  });
+  describe('withContext', () => {
+    it('creates a child context', () => {
+      const parent = new Confidence({} as any);
+      const additionalContext = {
+        clothes: 'pants',
+      };
+      let child = parent.withContext(additionalContext);
+      expect(child.getContext()).toEqual(additionalContext);
+    });
+    it('merge contexts', () => {
+      const parent = new Confidence({} as any);
+      parent.setContext({
+        clothes: 'pants',
+      });
+      const child = parent.withContext({
+        pants: 'blue',
+      });
+      expect(child.getContext()).toEqual({
+        clothes: 'pants',
+        pants: 'blue',
+      });
+    });
+  });
+  describe('create', () => {
+    it('creates a new confidence object', () => {
+      const confidenceProvider = Confidence.create({
+        clientSecret: 'secret',
+        region: 'us',
+        baseUrl: 'https://www.spotify.com',
+        environment: 'client',
+        fetchImplementation: {} as any,
+        timeout: 10,
+      });
+      expect(confidenceProvider.getContext()).toEqual({});
+    });
+  });
 });
--- a/packages/sdk/src/Confidence.test.ts
+++ b/packages/sdk/src/Confidence.test.ts
@@ -36,7 +36,7 @@ describe('Confidence', () => {
       const additionalContext = {
         clothes: 'pants',
       };
-      let child = parent.withContext(additionalContext);
+      const child = parent.withContext(additionalContext);
       expect(child.getContext()).toEqual(additionalContext);
     });
     it('merge contexts', () => {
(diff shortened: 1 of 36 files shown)


packages/sdk/src/Confidence.test.ts
```diff
@@ -0,0 +1,86 @@
...
+      expect(confidence.getContext()).toEqual(newContext);
+    });
+  });
+  describe('withContext', () => {
+    it('creates a child context', () => {
+      const parent = new Confidence({} as any);
+      const additionalContext = {
+        clothes: 'pants',
+      };
+      let child = parent.withContext(additionalContext);
```
nicklasl, line 39:
```suggestion
      const child = parent.withContext(additionalContext);
```

packages/sdk/src/Confidence.test.ts
```diff
@@ -0,0 +1,86 @@
...
+      expect(child.getContext()).toEqual({});
+      expect(parent.getContext()).toEqual({
+        clothes: 'pants',
+      });
+      expect(Object.keys(child.getContext())).toEqual([]);
+      child.clearContext()
+      expect(child.getContext()).toEqual({
+        clothes: 'pants',
+      });
+    })
```
nicklasl, line 71: nice test 👍

packages/sdk/src/Confidence.test.ts
```diff
@@ -0,0 +1,86 @@
...
+    it('creates a new confidence object', () => {
+      const confidenceProvider = Confidence.create({
+        clientSecret: 'secret',
+        region: 'us',
+        baseUrl: 'https://www.spotify.com',
+        environment: 'client',
+        fetchImplementation: {} as any,
+        timeout: 10,
+      });
+      expect(confidenceProvider.getContext()).toEqual({});
```
nicklasl, line 83: weird name of const.
```suggestion
      const confidence = Confidence.create({
        clientSecret: 'secret',
        region: 'us',
        baseUrl: 'https://www.spotify.com',
        environment: 'client',
        fetchImplementation: {} as any,
        timeout: 10,
      });
      expect(confidence.getContext()).toEqual({});
```

packages/sdk/src/Confidence.ts
```diff
@@ -0,0 +1,148 @@
+import { FlagResolverClient, FlagResolution, ApplyManager } from './FlagResolverClient';
+import { EventSenderEngine } from './EventSenderEngine';
+import { Value } from './Value';
+import { EventSender } from './events';
+import { Context } from './context';
+
+export { FlagResolverClient, FlagResolution };
```
nicklasl, line 7: What does it mean that we export these types? Should we do that already?

packages/sdk/src/context.ts
```diff
@@ -0,0 +1,16 @@
...
+  getContext(): Context;
+  setContext(context: Context): void;
+  updateContextEntry<K extends string>(name: K, value: Context[K]): void;
+  removeContextEntry(name: string): void;
+  withContext(context: Context): Self;
+  clearContext(): void;
+}
+
+export interface Context extends Value.Struct {
+  openFeature?: Value.Struct & {
```
nicklasl, line 13: Not blocking at all for this initial PR but just as a note: 
We want to declare any semantic context keys here asap in order to risk collisions later on (collisions such as customers defining context entries that we later plan as semantic contexts).

packages/sdk/src/events.ts
```diff
@@ -0,0 +1,6 @@
+import { Value } from './Value';
+import { Contextual } from './context';
+
+export interface EventSender extends Contextual<EventSender> {
+  sendEvent(name: string, message?: Value): void;
```
nicklasl, line 5: message should be a Struct?

packages/sdk/src/EventSenderEngine.ts
```diff
@@ -1,8 +1,116 @@
...
+    } else {
+      if (this.flushTimeoutMilliseconds > 0) {
+        this.pendingFlush = setTimeout(() => {
+          this.flush();
+        }, this.flushTimeoutMilliseconds);
+      }
+    }
+  }
+
+  public flush(): Promise<boolean> {
```
nicklasl, line 94: what does it mean that this is function is `public` while `send` is not?

packages/sdk/src/EventSenderEngine.ts
```diff
@@ -0,0 +1,116 @@
...
+      eventTime: new Date().toISOString(),
+      payload: { ...message, ...context },
+    });
+    if (this.pendingFlush) {
+      clearTimeout(this.pendingFlush);
+    }
+    if (this.writeQueue.length >= this.maxBatchSize) {
+      this.flush();
+    } else {
+      if (this.flushTimeoutMilliseconds > 0) {
```
nicklasl, line 86: What does it mean when `flushTimeoutMilliseconds` <= 0?
That we only flush on batch size?

packages/sdk/src/EventSenderEngine.ts
```diff
@@ -0,0 +1,116 @@
...
+        }
+        return request;
+      })
+      .build(fetchImplementation);
+  }
+
+  send(context: Value.Struct, name: string, message?: Value.Struct): void {
+    this.writeQueue.push({
+      eventDefinition: `eventDefinitions/${name}`,
+      eventTime: new Date().toISOString(),
```
nicklasl, line 77: nit: Worth sending a clock interface to be able to unit test the time things?

packages/sdk/src/EventSenderEngine.ts
nickybondarenko: In this case, all negative and 0 should mean that we don't set a timeout and just flush when a batch is full.

packages/sdk/src/EventSenderEngine.ts
nickybondarenko: We can use fake timers to do that

nicklasl (approved):

PR #106, opened 2024-04-23
refactor: add message container to payload

fabriziodemaria:
Created new flag `js-sdk-e2e-tests` in the backend, due to some irreversible schema changes mistakenly added to `js-sdk-e2e-test`

### DIFF
--- a/packages/sdk/src/EventSenderEngine.ts
+++ b/packages/sdk/src/EventSenderEngine.ts
@@ -80,9 +80,9 @@ export class EventSenderEngine {
 
   send(context: Value.Struct, name: string, message?: Value.Struct): void {
     this.writeQueue.push({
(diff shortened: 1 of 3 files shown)


packages/sdk/src/EventSenderEngine.ts
```diff
@@ -80,9 +80,9 @@ export class EventSenderEngine {
 
   send(context: Value.Struct, name: string, message?: Value.Struct): void {
     this.writeQueue.push({
-      eventDefinition: `eventDefinitions/${name}`,
```
fabriziodemaria, line 83: Not a deal breaker, but in other platform we have the Client adding this prefix. Especially when storage is used, it's nice to save events without this backend-specific requirement, and have the Client code handle that

vahidlazio (approved):

PR #128, opened 2024-05-02
feat: resolve flags on context change

andreas-karlsson:
Mostly changes that make our OpenFeature web-provider re-resolve whenever the associated Confidence instance updates it's context.

- [ ] All tests are passing


packages/sdk/src/Confidence.ts
```diff
@@ -33,10 +34,29 @@ export class Confidence implements EventSender, Trackable {
...
+  private flagResolution?: FlagResolutionPromise;
 
   constructor(config: Configuration, parent?: Confidence) {
     this.config = config;
     this.parent = parent;
+    this.contextChanges = subject(observer => {
+      let parentSubscription: Closer | void;
+      if (parent) {
+        parentSubscription = parent.contextChanges(keys => {
+          observer(keys.filter(key => !this._context.has(key)));
```
andreas-karlsson, line 51: If all keys are filtered we shouldn't emit!

packages/sdk/src/Confidence.ts
```diff
@@ -73,17 +93,26 @@ export class Confidence implements EventSender, Trackable {
   }
 
   setContext(context: Context): void {
+    const current = this.getContext();
+    const changedKeys: string[] = [];
     for (const key of Object.keys(context)) {
-      this.updateContextEntry(key, context[key]);
+      if (Value.equal(current[key], context[key])) continue;
```
andreas-karlsson, line 99: I think (!) we might be able to do reference equals here.

packages/sdk/src/Confidence.ts
```diff
@@ -73,17 +93,26 @@ export class Confidence implements EventSender, Trackable {
...
+    if (this.contextChanged && changedKeys.length > 0) {
+      this.contextChanged(changedKeys);
     }
-  }
-
-  private updateContextEntry<K extends string>(name: K, value: Context[K]) {
-    this._context.set(name, Value.clone(value));
   }
 
   clearContext(): void {
```
andreas-karlsson, line 108: We should remove this function. It's never really useful and it complicates things.

packages/sdk/src/Confidence.ts
```diff
@@ -106,7 +135,15 @@ export class Confidence implements EventSender, Trackable {
    * @internal
    */
   resolve(flagNames: string[]): Promise<FlagResolution> {
-    return this.config.flagResolverClient.resolve(this.getContext(), flagNames);
+    // we first resolve a promise so that if multiple context changes happen in the same tick, we still only make one resolve
+    return Promise.resolve().then(() => {
+      const context = this.getContext();
+      if (!this.flagResolution || !Value.equal(context, this.flagResolution.context)) {
+        this.flagResolution?.abort();
```
andreas-karlsson, line 142: Should put a reasonable reason in the abort call. Now it's just a generic "opeation was aborted" or something similar.

packages/sdk/src/context.ts
```diff
@@ -9,9 +9,7 @@ export interface Contextual<Self extends Contextual<Self>> {
 
 export interface Context extends Value.Struct {
   visitor_id?: string;
-  openFeature?: Value.Struct & {
-    targeting_key?: string;
-  };
+  targeting_key?: string;
```
andreas-karlsson, line 12: Do we even want semantic context for targeting_key? Do we ever expect that to be a thing in Confidence?

packages/sdk/src/observing.ts
```diff
@@ -0,0 +1,75 @@
...
+        pendingEmit = timeout(emit);
+      }
+      for (const value of values) {
+        buffer.add(value);
+      }
+    });
+    const closeEmit = () => pendingEmit?.();
+    return Closer.combine(closeEmit, closeSource);
+  };
+}
```
andreas-karlsson, line 75: We don't use this any longer, so could be ditched? But might become useful in the future...?

packages/sdk/src/Confidence.ts
andreas-karlsson: Or actually we won't. There's the case where the child set's the same value as the parent, that would trigger a spurious context change. Added a test for it.

packages/sdk/src/Confidence.ts
andreas-karlsson: Maybe later...

packages/sdk/src/context.ts
andreas-karlsson: For later...

packages/sdk/src/observing.ts
andreas-karlsson: Will remove!

nicklasl (approved):

PR #136, opened 2024-05-05
Rebuild build

andreas-karlsson:

This PR is a rewrite of our build process introducing the following changes:
- TS runs in project build mode, which is generally faster by an order of magnitude. It also make project wide refactoring easier, since for instance renames work across package boundaries.
- We generate an API report to keep track of the public API. Any updates to API reports need to be committed our the build will fail in CI.
- Sources and types are now bundled before release. Having each package only export one file makes it more portable. The bundling also strips internal types from the published type file.
- We use "yarn constraints" to make sure all package.json files are correctly setup, including that all intra dependencies are pointing to the correct version.

- [ ] All tests are passing


nicklasl (dismissed):

vahidlazio (approved):

yarn.config.cjs
```diff
@@ -0,0 +1,35 @@
...
+        continue;
+      }
+      workspace.set('type', 'module');
+      workspace.set('module', 'build/esm/index.js');
+      workspace.set('types', 'build/types/index.d.ts');
+      workspace.set('files', ['dist/index.*']);
+      workspace.set('scripts.build', 'tsc');
+      workspace.set(
+        'scripts.bundle',
+        '[ ${CI:-0} != 0 ] && api-extractor run --local || api-extractor run && rollup -c',
```
andreas-karlsson, line 24: @nicklasl I see you changed this. To me it looks broken? Shouldn't there be a parentheses around the api-extractor cases?

andreas-karlsson: Yeah, I just tried it, now it doesn't run rollup at all locally. Will fix.

nicklasl (approved):

PR #138, opened 2024-05-07
feat: resolve flags with confidence

andreas-karlsson:

- [ ] All tests are passing


.vscode/settings.json
```diff
@@ -1,4 +1,5 @@
 {
   "editor.defaultFormatter": "esbenp.prettier-vscode",
-  "editor.formatOnSave": true
+  "editor.formatOnSave": true,
+  "cmake.configureOnOpen": false
```
nicklasl, line 4: cmake?

packages/sdk/src/Value.ts
```diff
@@ -1,10 +1,44 @@
+import { TypeMismatchError } from './error';
+
 export namespace Value {
+  // TODO should lists be able to contain Structs?
```
nicklasl, line 4: no.

nicklasl (dismissed):

packages/sdk/src/flags.ts
```diff
@@ -52,7 +53,6 @@ export interface FlagResolver {
   // internal
   // evaluateFlagSync<T extends Value>(path: string, defaultValue: T): FlagEvaluation<T>;
 
-  evaluateFlag(path: string, defaultValue: number): FlagEvaluation<number>;
```
andreas-karlsson: @nicklasl we're actually gonna need these overloads for simple types, that's what saves us from getting the exact value inferred. Like `const v = c.evaluateFlag('a.b', 3)` Then v should have the type number rather than `3`.

nicklasl: ack!

andreas-karlsson:
Closed in favour of #141

PR #167, opened 2024-06-27
feat: improved suspense management

andreas-karlsson:

- Stop subscribing in useConfidence
- Subscribe to contextChanges in all hooks that depend on context (useFlagXXX, useContext)
- Handle stale flags through React transitions

- [ ] All tests are passing

### DIFF
--- a/packages/react/README.md
+++ b/packages/react/README.md
@@ -39,9 +39,11 @@ function App() {
 }
 ```
 
+Anywhere in the sub-tree under the `ConfidenceProvider` you can now access the confidence instance with the `useConfidence()` hook. The hook actually returns an instance of `ConfidenceReact` which is a wrapper around the normal `Confidence` API with some slight adaptations to integrate better with React. You can read more about the differences in the following sections.
+
 ### Managing context
 
-The `useConfidence()` hook supports the [standard context API's](https://github.com/spotify/confidence-sdk-js/blob/main/packages/sdk/README.md#setting-the-context). Additionally, the following wrapper component can be used to wrap a sub tree with additional context data.
+The `ConfidenceReact` instance supports the [standard context API](https://github.com/spotify/confidence-sdk-js/blob/main/packages/sdk/README.md#setting-the-context). Additionally, the following wrapper component can be used to wrap a sub tree with additional context data.
 
 ```ts
 <ConfidenceProvider.WithContext context={{ user_name: 'John Doe' }}>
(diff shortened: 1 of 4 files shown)


packages/react/src/index.tsx
```diff
@@ -146,26 +164,33 @@ export function useWithContext(context: Context, parent = useConfidence()): Conf
...
-        if (state === 'READY' || state === 'ERROR') setState(child.contextState);
-      }),
-    [child, setState],
-  );
   return child;
 }
 
+// this would be better named useContext, but would then collide with React.useContext
+// eslint-disable-next-line react-hooks/rules-of-hooks
+export function useConfidenceContext(confidence = useConfidence()): Context {
```
andreas-karlsson, line 172: I'm considering if we should put all the hooks as static methods on ConfidenceReact. Then this would become `ConfidenceReact.useContext` and it wouldn't collide with the `React.useContext` hook...

packages/react/README.md
```diff
@@ -54,11 +56,15 @@ The `useConfidence()` hook supports the [standard context API's](https://github.
...
 
-Both of the flag hooks integrate with the React Suspense API so that the suspense fallback will be visible until flag values are available. It is therefore important to wrap .
+The hooks are also reactive so that if the context changes, any components using the hooks will be re-rendered. As dependent components are re-rendered as soon as the context changes, one might expect that would again trigger the suspense boundary while the flag values are resolved. That is normally not the case however as the `ConfidenceReact.setContext` method is by default wrapped in a React [Transition](https://react.dev/reference/react/startTransition). If you would rather manage the transition logic yourself (with for example [React.useTransition()](https://react.dev/reference/react/useTransition)), or if you want to always trigger a suspense fallback to never show stale values, you can turn off the default behavior by calling `ConfidenceReact.setContext({...}, { transition:false })`.
```
nicklasl, line 65: 💯

packages/react/src/index.tsx
```diff
@@ -39,7 +48,7 @@ export class ConfidenceReact implements EventSender, Trackable, FlagResolver {
 
   /** @internal */
   get contextState(): string {
-    return this.delegate.flagState + Value.serialize(this.delegate.getContext());
```
nicklasl, line 42: Was this the only use of `flagState`? Maybe we can remove it from the API in that case?

packages/react/src/index.tsx
nicklasl: Not sure what the standard is... `ConfidenceReact.useContext` does feel a bit strange to me... had it been `Confidence.useContext` would maybe be better but also a bit strange IMO.

nicklasl (approved):
Nit about removing the `getFlagState()`

packages/react/src/index.tsx
```diff
@@ -146,26 +164,33 @@ export function useWithContext(context: Context, parent = useConfidence()): Conf
...
   const evaluation = confidence.delegate.evaluateFlag(path, defaultValue);
-  // TODO make it a setting to _enable skip throwing_ on stale value.
-  if (evaluation.reason === 'ERROR' && evaluation.errorCode === 'NOT_READY' && 'then' in evaluation) throw evaluation;
+  const [, setState] = useState(() => confidence.contextState);
+  useEffect(() => {
+    return confidence.delegate.contextChanges(() => {
+      setState(confidence.contextState);
+    });
+  });
+  if ('then' in evaluation) throw evaluation;
```
andreas-karlsson, line 193: We might have a slight problem here. Since we don't kick off a resolution until React calls the "then" method, we risk a "waterfall" loading behaviour. I.e. there are two sibling components that both are stale and need to fetch. The first component renders and throws a suspense. Then the second component wont be rendered until the first promise resolves and at that point the second component will throw and initiate a fetch... Actually, this shouldn't be the case in React 18 (where the second component will render to cache a potential suspense), but there was just rather big controversy about React 19 changing this behaviour, but then last minute they delayed React 19 due to backlash from the community. (See https://tkdodo.eu/blog/react-19-and-suspense-a-drama-in-3-acts) All that said, I think we'd be better off keeping track of which ConfidenceReact instances are currently serving flags to the UI and proactively initiate fetches for them when the context changes. Will have a quick look at this!

andreas-karlsson: Fixed.

packages/react/src/index.tsx
andreas-karlsson: Let's skip this for now.

nicklasl (approved):

PR #169, opened 2024-07-02
fix: shared requests aborted

andreas-karlsson:

Due to the client cache resolve requests can be shared. We keep a ref-count in the cache entries and clear the entry when the ref-count drops to zero. However we inadvertently aborted the original request as soon as any ref was aborted. Added tests to show this and a fix.

- [ ] All tests are passing

### DIFF
--- a/packages/sdk/src/FlagResolverClient.test.ts
+++ b/packages/sdk/src/FlagResolverClient.test.ts
@@ -1,8 +1,15 @@
-import { FetchingFlagResolverClient, withRequestLogic } from './FlagResolverClient';
+import {
+  CachingFlagResolverClient,
+  FetchingFlagResolverClient,
+  FlagResolverClient,
+  PendingResolution,
+  withRequestLogic,
+} from './FlagResolverClient';
 import { setMaxListeners } from 'node:events';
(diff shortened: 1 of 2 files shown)


packages/sdk/src/FlagResolverClient.test.ts
```diff
@@ -375,6 +382,40 @@ describe('withRequestLogic', () => {
...
+
+  it('should abort the original resolution when all references are aborted', () => {
+    const context = { a: 0 };
+    const pendingResolution = PendingResolution.create(context, () => new Promise(() => {}));
+    const cachingClient = new CachingFlagResolverClient(flagResolverClientMock, 1000);
+    flagResolverClientMock.resolve.mockReturnValue(pendingResolution);
+    const firstResolution = cachingClient.resolve(context, []);
+    const secondResolution = cachingClient.resolve(context, []);
+    firstResolution.abort();
+    firstResolution.abort();
```
nicklasl, line 413: Only need to be called once.

nicklasl (approved):

fabriziodemaria (approved):
I don't fully understand the logic changes, but tests look good 👍

PR #171, opened 2024-07-05
feat: support evict the flag cache

nicklasl:
This PR adds hacky support for manually evicting the Flag Cache if necessary.

### DIFF
--- a/packages/sdk/src/Confidence.ts
+++ b/packages/sdk/src/Confidence.ts
@@ -179,6 +179,15 @@ export class Confidence implements EventSender, Trackable, FlagResolver {
   }
 
   /**
+   * Evicts flag cache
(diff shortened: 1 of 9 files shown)


packages/sdk/src/Confidence.ts
```diff
@@ -147,6 +148,12 @@ export class Confidence implements EventSender, Trackable, FlagResolver {
     }
   }
 
+  clearFlagCache(): void {
+    if (this.config.flagResolverClient instanceof CachingFlagResolverClient) {
+      (this.config.flagResolverClient as CachingFlagResolverClient).forceEvict();
```
nicklasl, line 153: Not super elegant. We should consider this on the interface.

PR #174, opened 2024-07-10
feat: add support for common-js module

nicklasl:
With this PR we're additionally building out a "common-js" module to natively support legacy CommonJS projects.


andreas-karlsson (approved):
LGTM!

PR #175, opened 2024-08-07
fix: wip: hash secret in visitor id

fabriziodemaria:

- [ ] All tests are passing

### DIFF
--- a/packages/sdk/package.json
+++ b/packages/sdk/package.json
@@ -20,6 +20,7 @@
     "main": "dist/index.js"
   },
   "dependencies": {
+    "crypto-js": "^4.2.0",
     "web-vitals": "^3.5.2"
   },
   "files": [
@@ -27,6 +28,7 @@
   ],
   "devDependencies": {
     "@microsoft/api-extractor": "7.43.1",
+    "@types/crypto-js": "^4",
     "prettier": "*",
     "rollup": "4.14.2",
     "ts-proto": "^1.171.0"
--- a/packages/sdk/src/trackers/visitorId.ts
+++ b/packages/sdk/src/trackers/visitorId.ts
@@ -1,9 +1,10 @@
 import { Trackable } from '../Trackable';
 import { uuid, Cookie } from '../utils';
+import * as CryptoJS from 'crypto-js';
 
 const COOKIE_NAME = 'cnfdVisitorId';
 
-export const visitorIdentity = (): Trackable.Manager => controller => {
+export const visitorIdentity = (clientSecret: string): Trackable.Manager => controller => {
   if (typeof document === 'undefined') return;
   let value = Cookie.get(COOKIE_NAME);
   if (!value) {
@@ -11,5 +12,8 @@ export const visitorIdentity = (): Trackable.Manager => controller => {
     // TODO check correct cookie options
     Cookie.set(COOKIE_NAME, value, { maxAge: 60 * 60 * 24 * 365 * 5 });
   }
-  controller.setContext({ visitor_id: value });
+
+  const combinedString = value + clientSecret;
+  const hashedValue = CryptoJS.SHA256(combinedString).toString();
+  controller.setContext({ visitor_id: hashedValue });
 };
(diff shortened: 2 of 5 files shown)


packages/sdk/package.json
```diff
@@ -20,13 +20,15 @@
     "main": "dist/index.js"
   },
   "dependencies": {
+    "crypto-js": "^4.2.0",
```
nicklasl, line 23: This dep [weighs in at 487kB](https://www.npmjs.com/package/crypto-js). Not sure if you looked at any built in support? There seems to be some [built in support for browsers](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest#converting_a_digest_to_a_hex_string) given that the app runs on https. We should probably look into this or lightweight alternatives.

nicklasl (commented):
I like the solution in general but I think we need to investigate other ways of hashing.

package.json
```diff
@@ -12,7 +12,7 @@
     "format": "prettier --config prettier.config.js -w .",
     "format:check": "prettier --config prettier.config.js -c .",
     "lint": "eslint 'packages/*/src/**/*.{tsx,ts}'",
-    "test": "jest",
+    "test": "jest --testPathIgnorePatterns='packages/sdk/src/VisitorIdentifier.test.ts' && jest packages/sdk/src/VisitorIdentifier.test.ts --env=jsdom",
```
fabriziodemaria, line 15: Couldn't achieve this with test annotations or new configurations in `package.json`. This CLI-level configuration works, but it's not ideal

packages/openfeature-web-provider/src/ConfidenceWebProvider.e2e.test.ts
```diff
@@ -33,7 +33,7 @@ describe('ConfidenceWebProvider E2E tests', () => {
         errorCode: 'GENERAL',
         flagKey: 'web-sdk-e2e-flag.str',
         flagMetadata: {},
-        errorMessage: 'Resolve timeout',
+        errorMessage: '',
```
fabriziodemaria, line 36: Strange that this was required

package.json
nicklasl: So this is for allowing some type of tests to run in the `jsdom` environment?
Could we perhaps define an actual pattern for this? 
something like 
```
"test": "jest --testPathIgnorePatterns='packages/sdk/src/*.jsdom.test.ts' && jest packages/sdk/src/*.jsdom.test.ts --env=jsdom",
```
or however that would work?

packages/sdk/src/trackers/visitorId.ts
```diff
@@ -1,15 +1,28 @@
...
+    if (typeof document === 'undefined') return;
+    let value = Cookie.get(COOKIE_NAME);
+    if (!value) {
+      value = uuid();
+      // TODO check correct cookie options
+      Cookie.set(COOKIE_NAME, value, { maxAge: 60 * 60 * 24 * 365 * 5 });
+    }
+
+    const hashedValue = generateUniqueId(value + clientSecret).toString();
+    controller.setContext({ visitor_id: hashedValue });
```
nicklasl, line 18: Let's put the hashed value as the cookie value so we have the same visitor id stored in the cookie and in the context.

This will enable us to guide the customer to access the Cookie value from the requests and thus make it easier to use the same visitor ID both backend and frontend resolutions.

package.json
nicklasl: Really glad you took the time to figure out how to write these tests though! 👍 💯

packages/sdk/src/trackers/visitorId.ts
nicklasl, line 18:
```suggestion
      value = generateUniqueId(uuid() + clientSecret).toString();
      // TODO check correct cookie options
      Cookie.set(COOKIE_NAME, value, { maxAge: 60 * 60 * 24 * 365 * 5 });
    }

    controller.setContext({ visitor_id: value });
```

PR #176, opened 2024-08-08
fix: build issues

fabriziodemaria:
This fix unbreaks the `main` branch but we believe that this situation will occur again due to how release please works.

### DIFF
--- a/packages/openfeature-server-provider/package.json
+++ b/packages/openfeature-server-provider/package.json
@@ -8,12 +8,12 @@
     "@microsoft/api-extractor": "7.43.1",
     "@openfeature/core": "^1.1.0",
     "@openfeature/server-sdk": "^1.13.5",
-    "@spotify-confidence/sdk": "0.1.4",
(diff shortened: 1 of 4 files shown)


nicklasl (approved):

packages/openfeature-server-provider/package.json
```diff
@@ -8,12 +8,12 @@
...
     "@openfeature/core": "^1.1.0",
     "@openfeature/server-sdk": "^1.13.5",
-    "@spotify-confidence/sdk": "0.1.4",
+    "@spotify-confidence/sdk": "0.1.5",
     "rollup": "4.14.2"
   },
   "peerDependencies": {
     "@openfeature/server-sdk": "^1.13.5",
-    "@spotify-confidence/sdk": "^0.1.4"
+    "@spotify-confidence/sdk": "^0.1.5"
```
nicklasl, line 16: In theory, the server provider should be able to stay on `0.1.4` but if I change that locally and run `bundle`, the constraints tool gets sad.

PR #189, opened 2024-08-19
refactor: Standardise example apps

nickybondarenko:
This PR aligns the flags used in the example apps with the flags created during the connect guide/onboardling in the Confidence console.
Additionally, it removes hardcoded ClientSecret from the example apps using dotenv.


package.json
```diff
@@ -60,5 +60,8 @@
       "path": "@commitlint/cz-commitlint"
     }
   },
-  "packageManager": "yarn@4.1.0+sha256.81a00df816059803e6b5148acf03ce313cad36b7f6e5af6efa040a15981a6ffb"
+  "packageManager": "yarn@4.1.0+sha256.81a00df816059803e6b5148acf03ce313cad36b7f6e5af6efa040a15981a6ffb",
+  "dependencies": {
+    "dotenv": "^16.4.5"
+  }
```
nicklasl, line 66: Did you try to add this in devDependencies? Or maybe we can have it in every example-app package.json?

nickybondarenko: I brew-installed this. I'll try to remove since I _think_ we already have it

examples/react18/src/App.tsx
```diff
@@ -15,8 +15,13 @@ const handleFailRequestsOnChange = (e: React.ChangeEvent<HTMLInputElement>) => {
   state.failRequests = e.target.checked;
 };
 
+if (!process.env.REACT_APP_CLIENT_SECRET) {
+  console.error('REACT_APP_CLIENT_SECRET not set in .env');
```
nicklasl, line 19: Are we considering something similar for the other example apps?

examples/react18/src/App.tsx
```diff
@@ -15,8 +15,13 @@ const handleFailRequestsOnChange = (e: React.ChangeEvent<HTMLInputElement>) => {
...
+  process.exit(1);
+}
+
 const confidence = Confidence.create({
-  clientSecret: 'RxDVTrXvc6op1XxiQ4OaR31dKbJ39aYV',
+  clientSecret: process.env.REACT_APP_CLIENT_SECRET,
```
nicklasl, line 24: I suggest aligning on the same environment variable for all example apps.

nickybondarenko: React picks up only env variables with REACT_APP at the front, so we would be using the REACT_APP one, which can be confusing

examples/react18/src/App.tsx
nickybondarenko: I can add it!

examples/react18/src/App.tsx
nicklasl: That is really confusing! :shaking-fist:

nicklasl (approved):

PR #190, opened 2024-08-20
ci: run CI build on release prs

nicklasl:
This changes the Release amend to be run via a workflow dispatch instead of repository dispatch. 
It also adds a collection of return codes from a bunch of yarn commands that will then be added as a comment to the Release please open PR:

### DIFF
--- a/.github/workflows/Release-Please.yml
+++ b/.github/workflows/Release-Please.yml
@@ -29,8 +29,8 @@ jobs:
           curl -X POST \
           -H "Accept: application/vnd.github.v3+json" \
           -H "Authorization: token ${{ secrets.GITHUB_TOKEN }}" \
-          https://api.github.com/repos/spotify/confidence-sdk-js/dispatches \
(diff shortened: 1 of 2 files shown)


.github/workflows/Release-Please.yml
```diff
@@ -29,8 +29,8 @@ jobs:
           curl -X POST \
           -H "Accept: application/vnd.github.v3+json" \
           -H "Authorization: token ${{ secrets.GITHUB_TOKEN }}" \
-          https://api.github.com/repos/spotify/confidence-sdk-js/dispatches \
-          -d '{"event_type":"release_amend_trigger", "client_payload": {}}'
+          https://api.github.com/repos/spotify/confidence-sdk-js/actions/workflows/Amend-Release.yml/dispatches \
+          -d '{"ref":"release-please--branches--main", "inputs": {"branch": "release-please--branches--main"}}'
```
nicklasl, line 33: maybe ref should be set to `main`?
AFAIK this is which "version" of the workflow should be run and I assume we want the one from main to run?

PR #199, opened 2024-09-05
fix!: visitor id disabled by default

nicklasl:
We want to allow users to opt out from `visitor_id` which is a feature that uses cookies as a storage medium.
Since we are defaulting to disable visitor id (whereas it was previously enabled), we mark the commit as breaking.


andreas-karlsson (commented):
Would be nice to add some docs about this in at the same time?

PR #200, opened 2024-09-10
feat: add a custom baseURL for sidecar resolves

nickybondarenko:
With this PR, we're adding a possibility to use a custom url in order to sidecar resolves.

### DIFF
--- a/api/sdk.api.md
+++ b/api/sdk.api.md
@@ -50,15 +50,17 @@ export class Confidence implements EventSender, Trackable, FlagResolver {
 }
 
 // @public
-export type ConfidenceOptions = {
+export interface ConfidenceOptions {
     clientSecret: string;
-    region?: 'eu' | 'us';
     environment: 'client' | 'backend';
+    // Warning: (ae-forgotten-export) The symbol "SimpleFetch" needs to be exported by the entry point index.d.ts
     fetchImplementation?: SimpleFetch;
-    timeout: number;
+    // Warning: (ae-forgotten-export) The symbol "Logger" needs to be exported by the entry point index.d.ts
     logger?: Logger;
+    region?: 'eu' | 'us';
     resolveBaseUrl?: string;
-};
+    timeout: number;
+}
 
 // @public
 export interface Configuration {
@@ -212,11 +214,6 @@ export type WebVitalsOptions = {
     cls?: boolean;
     ttfb?: boolean;
 };
-
-// Warnings were encountered during analysis:
-//
-// src/Confidence.ts:34:3 - (ae-forgotten-export) The symbol "SimpleFetch" needs to be exported by the entry point index.d.ts
-// src/Confidence.ts:38:3 - (ae-forgotten-export) The symbol "Logger" needs to be exported by the entry point index.d.ts
 
 // (No @packageDocumentation comment for this package)
 
--- a/api/sdk.api.md
+++ b/api/sdk.api.md
@@ -22,7 +22,7 @@ export class Confidence implements EventSender, Trackable, FlagResolver {
     //
     // @internal
     readonly contextChanges: Subscribe<string[]>;
-    static create({ clientSecret, region, timeout, environment, fetchImplementation, logger, }: ConfidenceOptions): Confidence;
+    static create({ clientSecret, region, timeout, environment, fetchImplementation, logger, resolveBaseUrl, }: ConfidenceOptions): Confidence;
     get environment(): string;
     evaluateFlag(path: string, defaultValue: string): FlagEvaluation<string>;
     // (undocumented)
@@ -50,17 +50,15 @@ export class Confidence implements EventSender, Trackable, FlagResolver {
 }
 
 // @public
-export interface ConfidenceOptions {
+export type ConfidenceOptions = {
     clientSecret: string;
-    environment: 'client' | 'backend';
-    // Warning: (ae-forgotten-export) The symbol "SimpleFetch" needs to be exported by the entry point index.d.ts
-    fetchImplementation?: SimpleFetch;
-    // Warning: (ae-forgotten-export) The symbol "Logger" needs to be exported by the entry point index.d.ts
-    logger?: Logger;
     region?: 'eu' | 'us';
-    resolveUrl?: string;
+    environment: 'client' | 'backend';
+    fetchImplementation?: SimpleFetch;
     timeout: number;
-}
+    logger?: Logger;
+    resolveBaseUrl?: string;
+};
 
 // @public
 export interface Configuration {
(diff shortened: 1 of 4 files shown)


packages/sdk/src/Confidence.ts
```diff
@@ -38,6 +38,8 @@ export interface ConfidenceOptions {
   timeout: number;
   /** Debug logger */
   logger?: Logger;
+  /** Backend only option. Sets an alternative resolve url */
+  APIResolveBaseUrl?: string;
```
nicklasl, line 42: I would suggest
```suggestion
  customResolveBaseUrl?: string;
```

packages/sdk/src/FlagResolverClient.ts
```diff
@@ -103,13 +107,18 @@ export class FetchingFlagResolverClient implements FlagResolverClient {
...
     region,
+    APIResolveBaseUrl,
   }: FlagResolverClientOptions) {
     // TODO think about both resolve and apply request logic for backends
     this.fetchImplementation = environment === 'backend' ? fetchImplementation : withRequestLogic(fetchImplementation);
     this.clientSecret = clientSecret;
     this.sdk = sdk;
     this.applyTimeout = applyTimeout;
-    this.baseUrl = region ? `https://resolver.${region}.confidence.dev/v1` : 'https://resolver.confidence.dev/v1';
+    if (environment === 'backend' && APIResolveBaseUrl) {
```
nicklasl, line 117: I don't think this necessarily need to be only for backend.

packages/sdk/src/FlagResolverClient.ts
```diff
@@ -103,13 +107,18 @@ export class FetchingFlagResolverClient implements FlagResolverClient {
...
+      this.baseUrl = APIResolveBaseUrl;
```
nicklasl, line 118: We should append `/v1`.

packages/sdk/src/Confidence.ts
nicklasl, line 42: @andreas-karlsson is there some typescript way where we can make it visible in the API that `region` will not be respected if one provides the `customResolveBaseUrl`?

packages/sdk/src/FlagResolverClient.ts
nickybondarenko: good point!

packages/sdk/src/FlagResolverClient.ts
```diff
@@ -103,13 +104,18 @@ export class FetchingFlagResolverClient implements FlagResolverClient {
...
     region,
+    customResolveBaseUrl,
   }: FlagResolverClientOptions) {
     // TODO think about both resolve and apply request logic for backends
     this.fetchImplementation = environment === 'backend' ? fetchImplementation : withRequestLogic(fetchImplementation);
     this.clientSecret = clientSecret;
     this.sdk = sdk;
     this.applyTimeout = applyTimeout;
-    this.baseUrl = region ? `https://resolver.${region}.confidence.dev/v1` : 'https://resolver.confidence.dev/v1';
+    if (environment === 'backend' && customResolveBaseUrl) {
```
andreas-karlsson, line 114: Do we need to check for environment here? I think it might be interesting to allow setting a custom url also for clients.

packages/sdk/src/FlagResolverClient.ts
```diff
@@ -103,13 +104,18 @@ export class FetchingFlagResolverClient implements FlagResolverClient {
...
+      this.baseUrl = customResolveBaseUrl.concat('/v1');
```
andreas-karlsson, line 115: Use a template string instead. The concat method is never used in JS. Concat is usually done with the plus operator, but if you do that the linter will instead turn it into a template string which is the de facto standard for all string operations now days.
```suggestion
      this.baseUrl = `${customResolveBaseUrl}/v1`;
```

api/sdk.api.md
```diff
@@ -52,6 +52,7 @@ export class Confidence implements EventSender, Trackable, FlagResolver {
 // @public
 export interface ConfidenceOptions {
     clientSecret: string;
+    customResolveBaseUrl?: string;
```
andreas-karlsson, line 55: Could we change the option to `resolverBaseUrl`? I'm thinking "custom" is a little redundant in this context. And resolve vs resolver, I'm not sure, but "resolve" is the action, whereas resolver more signifies the service/resource which I think feels more appropriate..

packages/sdk/src/Confidence.ts
nicklasl: Nice! And that wouldn't even be a breaking change to the API, right? Because it unpacks the types?

packages/sdk/src/FlagResolverClient.ts
nickybondarenko: Oh so this is how it's better done 😮

api/sdk.api.md
nickybondarenko: Is it redundant though? My thinking is it's "custom" because you might want to specify it, but if you don't we use a standard url 🤔

packages/sdk/src/Confidence.ts
nickybondarenko: That would be in the API itself?

api/sdk.api.md
```diff
@@ -49,18 +49,16 @@ export class Confidence implements EventSender, Trackable, FlagResolver {
...
-    // Warning: (ae-forgotten-export) The symbol "SimpleFetch" needs to be exported by the entry point index.d.ts
     fetchImplementation?: SimpleFetch;
-    // Warning: (ae-forgotten-export) The symbol "Logger" needs to be exported by the entry point index.d.ts
-    logger?: Logger;
-    region?: 'eu' | 'us';
-    resolveUrl?: string;
     timeout: number;
-}
+    logger?: Logger;
+} & UrlOptions;
```
andreas-karlsson, line 62: Let's actually revert this UrlOptions thing and just have it as two optional properties. I looked a bit deeper into the [TS theory about this](https://stackoverflow.com/questions/76232866/mutually-exclusive-optional-properties), and mutually exclusive properties isn't possible to express cleanly. But more importantly, the user might actually need to supply both! Region is for all endpoints resolve/apply/publish, whereas the new one is only for resolve.

packages/sdk/src/Confidence.ts
```diff
@@ -323,6 +323,7 @@ export class Confidence implements EventSender, Trackable, FlagResolver {
    * @param environment - can be either "client" or "backend"
    * @param fetchImplementation - fetch implementation
    * @param logger - debug logger
+   * @param customResolveBaseUrl - custom backend resolve URL
```
andreas-karlsson, line 326: Oops.. here the it has the old name.

andreas-karlsson (approved):

PR #208, opened 2024-11-19
feat: Telemetry header

nicklasl:
This PR appends resolve latency telemetry on the header as well as sets the API for adding other type of telemetry.


packages/sdk/src/Telemetry.ts
```diff
@@ -0,0 +1,84 @@
...
+      libraryVersion: tag.version,
+      traces: [],
+    });
+    return (value: number) => {
+      this.pushTrace(tag, value);
+    };
+  }
+
+  getSnapshot(): Monitoring | undefined {
+    if (this.disabled) {
```
nicklasl, line 77: This, combined with how the interceptor treats it, decided wether or not we add the telemetry header.

packages/sdk/src/FlagResolverClient.ts
```diff
@@ -134,13 +152,17 @@ export class FetchingFlagResolverClient implements FlagResolverClient {
         this.resolveTimeout,
         new ResolveError('TIMEOUT', 'Resolve timeout'),
       );
+      const start = new Date().getTime();
```
andreas-karlsson, line 155: Here it's a little nicer to do `Date.now()` which just gives a unix ms epoch. Date objects are generally avoided.

packages/sdk/src/FlagResolverClient.ts
```diff
@@ -275,6 +297,25 @@ export class CachingFlagResolverClient implements FlagResolverClient {
   }
 }
 
+export function withTelemetryData(
+  fetchImplementation: (request: Request) => Promise<Response>,
```
andreas-karlsson, line 301: Hmm.. now that we have these two orthogonal fetch modifiers, maybe it would be nicer to pass a FetchBuilder to this function and the other. It's a little nicer than composing them.. wdyt?

packages/sdk/src/Telemetry.ts
andreas-karlsson: It's probably a bad idea though if we have tracing disabled to still keep adding them in the arrays. It's a memory leak.

packages/sdk/proto/confidence/telemetry/v1/telemetry.proto
```diff
@@ -0,0 +1,50 @@
...
+  PLATFORM_KOTLIN = 2;
+  PLATFORM_SWIFT = 3;
+  PLATFORM_JS_WEB = 4;
+  PLATFORM_JS_SERVER = 5;
+  PLATFORM_PYTHON = 6;
+  PLATFORM_GO = 7;
+  PLATFORM_RUBY = 8;
+  PLATFORM_RUST = 9;
+  PLATFORM_FLUTTER_IOS = 10;
+  PLATFORM_FLUTTER_ANDROID = 11;
```
andreas-karlsson, line 17: Might as well remove all but JS?

andreas-karlsson (approved):

andreas-karlsson (approved):

PR #209, opened 2024-11-26
feat: Resolve Debug to redirect to UI tool on web

nickybondarenko:
This PR adds a link to logger which can be used to access resolve debug tool in console.

### DIFF
--- a/packages/sdk/src/Confidence.ts
+++ b/packages/sdk/src/Confidence.ts
@@ -319,9 +319,9 @@ export class Confidence implements EventSender, Trackable, FlagResolver {
 
   private showLoggerLink(flag: string, context: Context) {
     this.config.logger.info?.(
-      `Resolve debug: https://app.confidence.spotify.com/flags/resolver-test?client-key=${
+      `[Resolve debug] https://app.confidence.spotify.com/flags/resolver-test?client-key=${
         this.config.clientSecret
-      }&flag=flags/${flag}&context=${JSON.stringify(context)}`,
+      }&flag=flags/${flag}&context=${encodeURIComponent(JSON.stringify(context))}`,
     );
   }
 
@@ -402,7 +402,7 @@ function defaultFetchImplementation(): typeof fetch {
 function defaultLogger(): Logger {
   try {
     if (process.env.NODE_ENV === 'development') {
-      return Logger.withLevel(console, 'info');
+      return Logger.withLevel(console, 'debug');
     }
   } catch (e) {
     // ignore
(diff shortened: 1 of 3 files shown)


packages/sdk/src/Confidence.ts
```diff
@@ -56,6 +56,8 @@ export interface Configuration {
   /** Flag Resolver Client
    * @internal */
   readonly flagResolverClient: FlagResolverClient;
+  /* @internal */
+  readonly clientSecret: string;
```
nicklasl, line 60: is it possible to declare it `private`? (not sure it helps but still)

packages/sdk/src/Confidence.ts
```diff
@@ -314,6 +317,14 @@ export class Confidence implements EventSender, Trackable, FlagResolver {
     return (await this.evaluateFlag(path, defaultValue)).value;
   }
 
+  private showLoggerLink(flag: string, context: Context) {
+    this.config.logger.info?.(
```
nicklasl, line 321: This is a too high level IMO. 
Would suggest `debug` or maybe even `trace`

packages/sdk/src/Confidence.ts
nickybondarenko: I don't think it's possible on an interface

packages/sdk/src/Confidence.ts
nickybondarenko: Chatted with Andreas about it: lower in Confidence class we only put this into info on "dev" builds. If we change it to debug/trace, the current factory instance won't show any of the logs in an app. So the user should then know to create a Confidence instance with a logger in debug mode.

nicklasl: I understand, sounds good!

nicklasl (approved):

packages/sdk/src/Confidence.ts
```diff
@@ -390,7 +402,7 @@ function defaultFetchImplementation(): typeof fetch {
 function defaultLogger(): Logger {
   try {
     if (process.env.NODE_ENV === 'development') {
-      return Logger.withLevel(console, 'info');
+      return Logger.withLevel(console, 'debug');
```
nicklasl, line 405: Was this a conscious decision?

nickybondarenko: yes! this way we default to debug logger in our demo apps and can see the logs when testing

nickybondarenko: yeah that makes sense. I'll change it back to info

nicklasl (approved):

andreas-karlsson (approved):

PR #213, opened 2025-01-28
update peer deps

msvens:

- [ ] All tests are passing


andreas-karlsson (commented):
This would be a breaking change. The real fix is that we want to still allow 0.1.4, but just also include 0.2.x. We could put it as `>=0.1.4` that wouldn't be breaking, and then we should only bump it when we actually need something new from the sdk.

msvens:
Ok thanks. Will go ahead with that.

andreas-karlsson (approved):

PR #217, opened 2025-02-04
feat(example): example showing how to use next.js with confidence

msvens:
Added an example showing how to integrate Confidence using 'use client' directives with next.js


ouwargui:
Hey there, sorry for barging in, if this is not the right place for this question let me know and I'll delete it.

I'm exploring experimentation platforms for my company, and Confidence looks like a great fit for our use cases.

We're using Nextjs and wondering if it's possible to mix the server and React SDKs. Specifically, we have Nextjs API routes that act as entry points, where we conditionally redirect users based on experiment flags. We'd like to use Confidence for this logic while also leveraging the client SDK for UI changes.

Is this use case supported? Additionally, if Confidence supports server-side experimentation, does that extend to React Server Components as well?

Thanks!

PR #228, opened 2025-03-04
fix: avoid cloning requests in FetchBuilder

### LINKED ISSUE #221: Invalid/undefined client secret hangs instead of being rejected
Using `sdk`. A typo in an env meant that we didn't pass a secret to `Confidence.create`. This lead to timeouts, rather than more severe errors, which confused us. After figuring out what had happened, I tried passing an invalid secret instead, which had the same behaviour.

Additionally, setting a long timeout of 10s appears to make the entire process lock up instead of timing out, and `node` logging this warning:

```
(node:3019436) MaxListenersExceededWarning: Possible EventTarget memory leak detected. 11 abort listeners added to [AbortSignal]. Use events.setMaxListeners() to increase limit
    at [kNewListener] (node:internal/event_target:572:17)
    at [kNewListener] (node:internal/abort_controller:241:24)
    at EventTarget.addEventListener (node:internal/event_target:685:23)
    at Function.addAbortListener (node:events:1244:12)
    at Object.addAbortListener (node:internal/deps/undici/undici:699:25)
    at _Request.clone (node:internal/deps/undici/undici:6246:16)
    at <redacted>/node_modules/@spotify-confidence/sdk/dist/index.cjs.js:1141:41
    at Generator.next (<anonymous>)
    at asyncGeneratorStep$3 (<redacted>/node_modules/@spotify-confidence/sdk/dist/index.cjs.js:934:28)
    at _next (<redacted>/node_modules/@spotify-confidence/sdk/dist/index.cjs.js:952:17)
```

`sdk/dist/index.cjs.js:1141` is in a `retry` function, so I guess there's an infinite retry loop on auth failure? For short timeouts it gets terminated before getting out of hand, but setting a long timeout lets it build up?

## Expected Behavior

Log a warning when secret is missing/invalid. Preferably different

## Current Behavior

Didn't read the whole template before writing... See above

## Steps to Reproduce

```js
    clientSecret: process.env.CONFIDENCE_SECRET,
    environment: 'server',
    timeout: 10000,
    logger
  })
```

Where `process.env.CONFIDENCE_SECRET` is missing or invalid.

## Context

N/A

## Your Environment

- NodeJS Version: v20.11.1
- Operating System and Version: Linux 6.12.11-200.fc41.x86_64
- Browser Information: N/A

andreas-karlsson:

Minimal fix to circumvent the [this](https://github.com/nodejs/undici/issues/4068) bug in node which looses AbortSignals on cloned requests which in turn makes our retries infinite. The idea of the fix is to avoid creating requests in FetchBuilder until the very last step just before sending the request.

fixes #221 

- [ ] All tests are passing

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -19,6 +19,7 @@
     "es/no-optional-chaining": "off",
     "es/no-nullish-coalescing-operators": "off",
     "no-inner-declarations": "off",
-    "@typescript-eslint/no-use-before-define": "off"
+    "@typescript-eslint/no-use-before-define": "off",
+    "no-param-reassign": "off"
   }
 }
--- a/packages/sdk/src/fetch-util.ts
+++ b/packages/sdk/src/fetch-util.ts
@@ -1,8 +1,3 @@
-/**
- * Convenient type alias for the full fetch API
- */
-type Fetch = typeof fetch;
-
 export const enum TimeUnit {
   SECOND = 1000,
   MINUTE = 60 * TimeUnit.SECOND,
@@ -23,13 +18,7 @@ export const enum TimeUnit {
  * const fullFetch:Fetch = (input, init) => simple(new Request(input, init))
  * ```
  */
-type SimpleFetch = (request: Request) => Promise<Response>;
-
-/**
- * A function that takes a SimpleFetch and returns a new SimpleFetch
- * with some desired behavior applied.
- */
-type FetchPrimitive = (next: SimpleFetch) => SimpleFetch;
+export type SimpleFetch = (request: Request) => Promise<Response>;
 
 export class RequestError extends Error {}
 export class TimeoutError extends RequestError {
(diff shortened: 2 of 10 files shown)


.eslintrc
```diff
@@ -19,6 +19,7 @@
     "es/no-optional-chaining": "off",
     "es/no-nullish-coalescing-operators": "off",
     "no-inner-declarations": "off",
-    "@typescript-eslint/no-use-before-define": "off"
+    "@typescript-eslint/no-use-before-define": "off",
+    "no-param-reassign": "off"
```
andreas-karlsson, line 23: I think reassigning a parameter is fine. It's just a variable.

packages/sdk/src/fetch-util.ts
```diff
@@ -102,27 +103,12 @@ export class FetchBuilder {
...
         clearTimeout(timeoutId);
       });
     });
   }
 
-  /**
-   * Only allow one outstanding request. If a new request is made, the existing request it aborted with {@link RequestError}
-   *
-   * @returns the builder itself
-   */
```
nickybondarenko, line 115: Does this mean we will be allowing multiple requests at the same time?

andreas-karlsson: Yes. Or we actually weren't using this, so I opted for removing it. We have limitPending which can limit the number of concurrent requests.

PR #238, opened 2025-03-28
build: backport exports to support node moduleResolution

nicklasl:
We noticed that apps that use `compilerOptions.moduleResolution = node` (in the tsconfig) do not work well with exports.
This approach should backport to keep supporting legacy moduleResolution but also allow us to use exports for multiple entry points (which is used in the `react` package).

Also included here is that we restore the previous entry point for `react` to not break `imports`.


packages/react/package.json
```diff
@@ -3,12 +3,14 @@
...
   "files": [
-    "dist/*"
+    "dist/*",
+    "server/*"
   ],
   "scripts": {
     "bundle": "rollup -c && ../../validate-api.sh",
     "build": "tsc",
-    "prepack": "yarn build && yarn bundle",
+    "prepack": "yarn build && yarn bundle && mkdir -p ./server && mv dist/server.* ./server/",
```
andreas-karlsson, line 12: We need to either copy the server files to the package root, and then they can be named `server.*`, but if we copy them into a server folder they need to be named `index.*`.

It also bugs me a bit that the prepack/postpack "destroys" the bundle output. Maybe it'd be better copying the files and then removing.

nicklasl: fixed.

andreas-karlsson (dismissed):
Nice!

PR #239, opened 2025-04-01
fix: properly close the iterator in toOptions

andreas-karlsson:

- [ ] All tests are passing

### DIFF
--- a/packages/sdk/src/abstract-cache.ts
+++ b/packages/sdk/src/abstract-cache.ts
@@ -30,14 +30,6 @@ export abstract class AbstractCache<K, T, S = unknown> implements AsyncIterable<
     this.load(entries);
   }
 
-  ref() {
-    this.pendingUpdates++;
-  }
-
-  unref() {
-    this.pendingUpdates--;
-  }
-
   protected abstract serialize(value: T): S;
 
   protected abstract deserialize(data: S): T;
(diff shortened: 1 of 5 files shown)


packages/sdk/src/abstract-cache.ts
```diff
@@ -30,14 +30,6 @@ export abstract class AbstractCache<K, T, S = unknown> implements AsyncIterable<
     this.load(entries);
   }
 
-  ref() {
-    this.pendingUpdates++;
```
nicklasl, line 34: Were we doing a double `this.pendingUpdates++` previously? Since this happens in `put()` as well.

nicklasl: (and doulbe -- of course)

packages/sdk/src/abstract-cache.ts
```diff
@@ -48,15 +40,24 @@ export abstract class AbstractCache<K, T, S = unknown> implements AsyncIterable<
...
+  async *entries(keepOpen?: Promise<void>): AsyncIterator<[string, S]> {
+    keepOpen = keepOpen?.then(() => {
+      keepOpen = undefined;
+    });
     let cursor = this.head;
-    while (!('then' in cursor) || this.pendingUpdates) {
-      const node = await cursor;
-      yield [node.key, this.serialize(node.value)];
-      cursor = node.next;
+    while (!('then' in cursor) || this.pendingUpdates || keepOpen) {
```
nicklasl, line 48:
```suggestion
    const cursorHasPendingPromise = 'then' in cursor;
    while (!cursorHasPendingPromise || this.pendingUpdates || keepOpen) {
```
Not an actual suggestion but if we rewrite it like this, I'm still confused how to think about the `not`.

nicklasl (commented):
Is it possible to write tests for this?

andreas-karlsson:
Closed in favour of #240

PR #246, opened 2025-04-03
feat(react): react server support

andreas-karlsson:
BREAKING CHANGE:
Previously, when you did `useConfidence`, the returned instance would be augmented with extra hook-like functions such as `useFlag`. These have been removed and you should just use the exported top level hooks.

---------

### DIFF
--- a/packages/react/README.md
+++ b/packages/react/README.md
@@ -16,15 +16,17 @@ yarn add @spotify-confidence/react
 
 ### Initializing the ConfidenceProvider
 
-The Confidence React integration has a Provider that needs to be initialized. It accepts a Confidence instance and should wrap your component tree.
+The Confidence React integration has a Provider that needs to be initialized. It accepts a Confidence instance and should wrap your component tree. Here's an example for client-side rendering:
 
 ```ts
-import { Confidence } from '@spotify-confidence/sdk';
+import { Confidence } from '@spotify-confidence/sdk/';
+import { ConfidenceProvider } from '@spotify-confidence/sdk/react/client';
 
+// Client-side initialization
 const confidence = Confidence.create({
   clientSecret: 'mysecret',
   region: 'eu',
-  environment: 'client',
+  environment: 'client', // Note: For client-side rendering
   timeout: 1000,
 });
 
@@ -38,6 +40,8 @@ function App() {
   );
 }
 ```
+
+For server-side rendering setup, see the [Server-Side Rendering Support](#server-side-rendering-support) section below.
 
 Anywhere in the sub-tree under the `ConfidenceProvider` you can now access the confidence instance with the `useConfidence()` hook. The hook actually returns an instance of `ConfidenceReact` which is a wrapper around the normal `Confidence` API with some slight adaptations to integrate better with React. You can read more about the differences in the following sections.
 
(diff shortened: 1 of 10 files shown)


nicklasl (approved):

packages/react/README.md
```diff
@@ -16,15 +16,17 @@ yarn add @spotify-confidence/react
 
 ### Initializing the ConfidenceProvider
 
-The Confidence React integration has a Provider that needs to be initialized. It accepts a Confidence instance and should wrap your component tree.
+The Confidence React integration has a Provider that needs to be initialized. It accepts a Confidence instance and should wrap your component tree. Here's an example for client-side rendering:
 
 ```ts
-import { Confidence } from '@spotify-confidence/sdk';
+import { Confidence } from '@spotify-confidence/sdk/';
+import { ConfidenceProvider } from '@spotify-confidence/sdk/react/client';
```
andreas-karlsson, line 23:
```suggestion
import { ConfidenceProvider } from '@spotify-confidence/sdk/react';
```

packages/react/README.md
```diff
@@ -39,6 +41,8 @@ function App() {
 }
 ```
 
+For server-side rendering setup, see the [Server-Side Rendering Support](#server-side-rendering-support) section below.
+
 Anywhere in the sub-tree under the `ConfidenceProvider` you can now access the confidence instance with the `useConfidence()` hook. The hook actually returns an instance of `ConfidenceReact` which is a wrapper around the normal `Confidence` API with some slight adaptations to integrate better with React. You can read more about the differences in the following sections.
```
andreas-karlsson, line 46: > The hook actually returns an instance of `ConfidenceReact`

Not true anymore.

packages/react/README.md
```diff
@@ -66,6 +70,127 @@ The hooks are also reactive so that if the context changes, any components using
 
 If the hooks can't resolve the flag values withing the timeout specified on the Confidence instance, they will instead return the default value.
 
+### Server-Side Rendering Support (experimental)
+
+The Confidence React SDK now supports server-side rendering (SSR) and React Server Components (RSC) for instance in Next.js. The SDK provides separate entry points for client and server components:
+
+- `@spotify-confidence/react/client` - For client components
```
andreas-karlsson, line 77:
```suggestion
- `@spotify-confidence/react` - For client components
```

packages/react/README.md
```diff
@@ -66,6 +70,127 @@ The hooks are also reactive so that if the context changes, any components using
 
 If the hooks can't resolve the flag values withing the timeout specified on the Confidence instance, they will instead return the default value.
 
+### Server-Side Rendering Support (experimental)
+
+The Confidence React SDK now supports server-side rendering (SSR) and React Server Components (RSC) for instance in Next.js. The SDK provides separate entry points for client and server components:
```
andreas-karlsson, line 75:
```suggestion
The Confidence React SDK now supports server-side rendering (SSR) and React Server Components (RSC) for instance in Next.js. The SDK provides a separate entry point for server components:
```

packages/react/README.md
```diff
@@ -66,6 +70,127 @@ The hooks are also reactive so that if the context changes, any components using
...
+
+- `@spotify-confidence/react/client` - For client components
+- `@spotify-confidence/react/server` - For server components
+
+When using the SDK in a server environment:
+
+1. Create a Confidence instance for server using the React.cache as the scope in CacheOptions.
+2. Provide an accessor for the Confidence instance using `withContext`.
```
andreas-karlsson, line 83: This is a bit cryptic?

packages/react/README.md
```diff
@@ -66,6 +70,127 @@ The hooks are also reactive so that if the context changes, any components using
...
+  logger: console,
+  cache: {
+    scope: React.cache, // Use React.cache for server-side caching
+  },
+});
+
+// Confidence accessor for use in RSC.
+export const getConfidence = (context: Context): Confidence => {
+  return confidence.withContext(context);
+};
```
andreas-karlsson, line 106: We don't need this step. We could recommend to just export the root instance and then do withContext(), maybe that'd be easier to document?

packages/react/README.md
```diff
@@ -66,6 +70,127 @@ The hooks are also reactive so that if the context changes, any components using
```
nicklasl, line 70: @andreas-karlsson can you check [L69](https://github.com/spotify/confidence-sdk-js/pull/246/files#diff-3dae6d110b675ef311b837cf10653c25945e787593bfc47d7bdbc3836df15abbR69)?

packages/react/README.md
nicklasl: fixed

packages/react/README.md
```diff
@@ -58,14 +56,127 @@ Flags are accessed with a set of hooks exported from `@spotify-confidence/react`
...
 
-The hooks are also reactive so that if the context changes, any components using the hooks will be re-rendered. As dependent components are re-rendered as soon as the context changes, one might expect that would again trigger the suspense boundary while the flag values are resolved. That is normally not the case however as the `ConfidenceReact.setContext` method is by default wrapped in a React [Transition](https://react.dev/reference/react/startTransition). If you would rather manage the transition logic yourself (with for example [React.useTransition()](https://react.dev/reference/react/useTransition)), or if you want to always trigger a suspense fallback to never show stale values, you can turn off the default behavior by calling `ConfidenceReact.setContext({...}, { transition:false })`.
+The hooks are also reactive so that if the context changes, any components using the hooks will be re-rendered. As dependent components are re-rendered as soon as the context changes, one might expect that would again trigger the suspense boundary while the flag values are resolved.
```
andreas-karlsson, line 61: > one might expect that would again trigger the suspense boundary while the flag values are resolved.
That sentence without the followup is little strange 😄

nicklasl: I had AI rewrite the main part of the access flags. let me know what you think.

nicklasl (approved):

packages/react/README.md
```diff
@@ -39,32 +41,207 @@ function App() {
...
-  <UserDetails />
-</ConfidenceProvider.WithContext>
+import { useFlag, useEvaluateFlag } from '@spotify-confidence/react';
+
+function MyComponent() {
+  // Simple flag access - returns the flag value or default
+  const color = useFlag('my-feature-flag.color', 'blue');
+
+  // Detailed flag evaluation - returns evaluation details
+  const { value, reason, context } = useEvaluateFlag('my-feature-flag.size', 12);
```
nicklasl, line 64:
```suggestion
  const { value, reason } = useEvaluateFlag('my-feature-flag.size', 12);
```

nicklasl (approved):

PR #252, opened 2025-04-08
fix: update the resolver tester message to exclude link

nicklasl:
Resolver hint logs will now contain a base64 encoded string which can be pasted in the Confidence tools to support that we don't necessarily know the URL the user would want to use.

### DIFF
--- a/packages/sdk/src/Confidence.ts
+++ b/packages/sdk/src/Confidence.ts
@@ -15,6 +15,7 @@ import { AccessiblePromise } from './AccessiblePromise';
 import { Telemetry } from './Telemetry';
 import { SimpleFetch } from './fetch-util';
 import { CacheOptions, CacheProvider, FlagCache } from './flag-cache';
+import { utf8ToBase64 } from './utils';
 
 /**
  * Confidence options, to be used for easier initialization of Confidence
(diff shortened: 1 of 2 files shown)


packages/sdk/src/Confidence.ts
```diff
@@ -329,10 +330,16 @@ export class Confidence implements EventSender, Trackable, FlagResolver {
...
+      flag,
+      context,
+    };
+    const jsonString = JSON.stringify(data);
+    const base64 =
+      typeof window !== 'undefined' ? utf8ToBase64(jsonString) : Buffer.from(jsonString).toString('base64');
     this.config.logger.info?.(
-      `See resolves for '${flag}' in Confidence: https://app.confidence.spotify.com/flags/resolver-test?client-key=${
-        this.config.clientSecret
-      }&flag=flags/${flag}&context=${encodeURIComponent(JSON.stringify(context))}`,
```
fabriziodemaria, line 335: Did we lose the `flags/` prefix in the new encoded blob?

fabriziodemaria (changes requested):

packages/sdk/src/Confidence.ts
nicklasl: yes, thanks for catching this!

packages/sdk/src/Confidence.ts
```diff
@@ -329,10 +330,16 @@ export class Confidence implements EventSender, Trackable, FlagResolver {
   }
 
   private showLoggerLink(flag: string, context: Context) {
+    const data = {
+      clientKey: this.config.clientSecret,
+      flag: `flags/${flag}`,
```
nicklasl, line 335: Fixed up to prepend `flags/`

nicklasl: cc @fabriziodemaria

fabriziodemaria (dismissed):
LGTM

packages/sdk/src/utils.ts
```diff
@@ -37,6 +37,16 @@ export function uuid(): string {
   return 'xxxxxxxx-xxxx-4xxx-xxxx-xxxxxxxxxxxx'.replace(/x/g, () => HEX_ALPHA[(Math.random() * 16) | 0]);
 }
 
+export function utf8ToBase64(str: string): string {
+  const encoder = new TextEncoder();
+  const bytes = encoder.encode(str);
+  let binary = '';
+  for (let i = 0; i < bytes.length; i++) {
+    binary += String.fromCharCode(bytes[i]);
+  }
```
andreas-karlsson, line 46: This would be a bit more idiomatic and efficient:
```suggestion
  const binary = String.fromCharCode(...bytes);
```

andreas-karlsson (dismissed):

andreas-karlsson (approved):

PR #260, opened 2025-04-28
docs: add docs about logging

nicklasl:
Add part about logging

### DIFF
--- a/packages/sdk/README.md
+++ b/packages/sdk/README.md
@@ -38,6 +38,34 @@ The current regions are: `eu` and `us`.
 
 The timeout option is used to set the timeout for the feature flag resolve network request to the Confidence backend. When the timeout is reached, default values will be returned.
 
+### Logging
+
+During your integration and when debugging, you can get helpful logging information by defining a `logger` when creating the Confidence instance. The `Logger` is an interface for you to implement, or you can also pass `console`.
+
+```ts
+import { Confidence } from '@spotify-confidence/sdk';
+
+const myLogger = {
+  info: _ => {
+    // don't log anything
+  },
+  warn: message => {
+    console.log('Confidence warning: ', message);
+  },
+  error: message => {
+    console.log('Confidence error: ', message);
+  },
+};
+
+const confidence = Confidence.create({
+  clientSecret: 'mysecret',
+  region: 'eu',
+  environment: 'client',
+  logger: myLogger, // or console
+  timeout: 1000,
+});
+```
+
 ## Setting the context
 
 You can set the context manually by using `setContext({})`:
--- a/packages/sdk/README.md
+++ b/packages/sdk/README.md
@@ -40,15 +40,12 @@ The timeout option is used to set the timeout for the feature flag resolve netwo
 
 ### Logging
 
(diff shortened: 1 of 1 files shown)


packages/sdk/README.md
```diff
@@ -38,6 +38,34 @@ The current regions are: `eu` and `us`.
...
+  error: message => {
+    console.log('Confidence error: ', message);
+  },
+};
+
+const confidence = Confidence.create({
+  clientSecret: 'mysecret',
+  region: 'eu',
+  environment: 'client',
+  logger: myLogger, // or console
```
msvens, line 64: Would it make sense to also show how you pass 'console' or is that understandable? I.e. I interpret it as you would do logger: 'logger'

nicklasl: I was hope my comment `// or console` would be explanation enough.
Passing console would look like
```
...
environment: 'client',
logger: console,
...
```

msvens: Ah sorry missed the // or console. Then Im happy

packages/sdk/README.md
```diff
@@ -38,6 +38,34 @@ The current regions are: `eu` and `us`.
...
+
+During your integration and when debugging, you can get helpful logging information by defining a `logger` when creating the Confidence instance. The `Logger` is an interface for you to implement, or you can also pass `console`.
+
+```ts
+import { Confidence } from '@spotify-confidence/sdk';
+
+const myLogger = {
+  info: _ => {
+    // don't log anything
+  },
```
andreas-karlsson, line 51: This should instead just be removed. I.e. If you don't want to log info, don't provide info.
```suggestion
```

packages/sdk/README.md
```diff
@@ -38,6 +38,34 @@ The current regions are: `eu` and `us`.
 
 The timeout option is used to set the timeout for the feature flag resolve network request to the Confidence backend. When the timeout is reached, default values will be returned.
 
+### Logging
+
+During your integration and when debugging, you can get helpful logging information by defining a `logger` when creating the Confidence instance. The `Logger` is an interface for you to implement, or you can also pass `console`.
```
andreas-karlsson, line 43:
```suggestion
During your integration and when debugging, you can get helpful logging information by defining a `logger` when creating the Confidence instance. The `Logger` is an interface for you to implement. It's very similar to the console object, but all the logging functions (`debug`, `info`, `warn` etc) are optional, so you just provide the ones you are interested in. Providing console as the logger will log everything to the console. If you don't want any logging you can provide `{}` which is also a valid `Logger` implementation. If no logger is provided it will default to logging `info` or higher in development, but no logging in production.
```

packages/sdk/README.md
andreas-karlsson: Maybe we should remove the `// or console` comment. I don't think anyone would be interested in always logging everything?

andreas-karlsson: I think the most sensible thing is not set the logger at all, especially if we can get rid of the excessive resolve tester logging.

andreas-karlsson (approved):

PR #261, opened 2025-04-29
Use native AbortSignal.timeout (if exists)

andreas-karlsson:
I had to also update typescript, cause AbortSignal.any wasn't available in the standard lib types for the version we were using.

### DIFF
--- a/packages/sdk/src/signals.ts
+++ b/packages/sdk/src/signals.ts
@@ -1,0 +1,33 @@
+export namespace Signals {
+  const anyPolyfill: (typeof AbortSignal)['any'] = signals => {
+    const controller = new AbortController();
+    const abort = () => {
+      controller.abort();
+      clear();
+    };
+    const clear = () => {
+      signals.forEach(signal => {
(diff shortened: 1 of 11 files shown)


packages/sdk/src/signals.ts
```diff
@@ -0,0 +1,33 @@
...
+    return controller.signal;
+  };
+
+  const timeoutPolyfill: (typeof AbortSignal)['timeout'] = milliseconds => {
+    const controller = new AbortController();
+    setTimeout(() => {
+      controller.abort(new DOMException('signal timed out', 'TimeoutError'));
+    }, milliseconds);
+    return controller.signal;
+  };
```
nicklasl, line 29: This does not contain the "special stuff" we previously had for NodeJS. Will all node version contain the native `AbortSignal.timeout`?

this 
```
  // in Node setTimeout returns an object, with an unref function which will prevent the timeout from keeping the process alive
  if (typeof timeoutId === 'object') timeoutId.unref();
```

PR #262, opened 2025-05-07
fix: null Value treated as undefined, fix crashes

fabriziodemaria:
Null values in context were causing crashes


andreas-karlsson (approved):
Nice!

PR #265, opened 2025-05-15
fix: add support for resolveUrl in createConfidenceServerProvider()

nicklasl:
This will allow you to pass the `resolveBaseUrl` directly when using `createConfidenceServerProvider()`

### DIFF
--- a/api/openfeature-server-provider.d.ts
+++ b/api/openfeature-server-provider.d.ts
@@ -32,6 +32,8 @@ type ConfidenceProviderFactoryOptions = {
     fetchImplementation?: typeof fetch;
     clientSecret: string;
     timeout: number;
+    /** Sets an alternative resolve url */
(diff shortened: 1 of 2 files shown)


nickybondarenko:
Does the handing already exist in the code?

nicklasl:
> Does the handing already exist in the code?

Yes, since the [implementation of the function](https://github.com/spotify/confidence-sdk-js/blob/resolve-url-for-provider/packages/openfeature-server-provider/src/factory.ts#L38) is:
```ts
    ...confidenceOrOptions,
    environment: 'backend',
  });
```
and the param has the same name in [ConfidenceOptions](https://github.com/spotify/confidence-sdk-js/blob/resolve-url-for-provider/packages/sdk/src/Confidence.ts#L38) it will just work ™️ .

andreas-karlsson (approved):

PR #273, opened 2025-07-17
docs: restructure and update to docs

Lovisa01:

- Changed the README in the root folder to give a better overview of the different packages and provide a quick overview of how to use the SDK with OpenFeature in both a client- and a server-side environment. 
- Added docs of how to use the sdk in an SSR environment and link to example application 
- Fixed minor errors

- [ ] All tests are passing

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,14 +8,14 @@ JavaScript implementation of the [Confidence](https://confidence.spotify.com/) S
 
 This monorepo contains four packages:
 
-| Package | Description |
-|---------|----------|
-| [`openfeature-server-provider`](./packages/openfeature-server-provider) | OpenFeature provider for server-side environments | 
-| [`openfeature-web-provider`](./packages/openfeature-web-provider) | OpenFeature provider for client-side web applications | 
-| [`sdk`](./packages/sdk) | Core Confidence SDK for flag resolution, context, and tracking | 
-| [`react`](./packages/react) | React hooks and providers for client-side applications | 
+| Package                                                                 | Description                                                    |
+| ----------------------------------------------------------------------- | -------------------------------------------------------------- |
+| [`openfeature-server-provider`](./packages/openfeature-server-provider) | OpenFeature provider for server-side environments              |
+| [`openfeature-web-provider`](./packages/openfeature-web-provider)       | OpenFeature provider for client-side web applications          |
+| [`sdk`](./packages/sdk)                                                 | Core Confidence SDK for flag resolution, context, and tracking |
+| [`react`](./packages/react)                                             | React hooks and providers for client-side applications         |
 
-Read more about the packages in their respective docs. 
+Read more about the packages in their respective docs.
 
 ---
 
@@ -53,8 +53,8 @@ const client = OpenFeature.getClient();
 
 ```ts
 const isEnabled = await client.getBooleanValue('feature.enabled', false, {
-    targetingKey: `your-targeting-key`,
-  });
+  targetingKey: `your-targeting-key`,
+});
 ```
 
 > **Learn more**: [Server Provider Documentation](./packages/openfeature-server-provider/README.md)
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -4,98 +4,119 @@
...
 
-```ts
-const confidence = Confidence.create({...});
-const flag = confidence.withContext({}).getFlag('flag', {})
-```
+> **Learn more**: [SDK Documentation](./packages/sdk/README.md)
+
+---
+
+# React Integration
```
andreas-karlsson, line 107: Maybe this should be another level in since it only applies to Direct SDK Usage.

README.md
```diff
@@ -4,98 +4,119 @@
...
+
+For React applications, use the dedicated React package that provides hooks and providers for seamless integration.
+
+> **Learn more**: [React Integration Documentation](./packages/react/README.md)
+
+---
+
+# Example Application
```
andreas-karlsson, line 115: Maybe this should be the last sub heading under "Usage with OpenFeature" since it applies to that. Would also conform with moving everything OF related to the top.

README.md
```diff
@@ -4,98 +4,119 @@
...
+const client = OpenFeature.getClient();
+const isEnabled = client.getBooleanValue('feature.enabled', false);
 ```
 
-#### Caching
+> **Learn more**: [Web Provider Documentation](./packages/openfeature-web-provider/README.md)
 
-Flag evaluations are cached in memory on the Confidence instance with the evaluation context and flag name as a cache key.
-This is done to reduce network calls when evaluating multiple flags using the same context.
+---
```
andreas-karlsson, line 97: Maybe we could also add a note here that OpenFeature has a React SDK.

packages/react/README.md
```diff
@@ -4,23 +4,23 @@
 
 This package contains helper functionality to make the Confidence SDK work well in a React environment.
```
andreas-karlsson, line 5: Maybe add a comment that this react package is only relevant if using the Confidence SDK directly and that if using OpenFeature there is an OpenFeature React SDK.

packages/react/README.md
```diff
@@ -128,21 +128,18 @@ function MyComponent() {
    const name: string = useFlag('user.name', '');
    ```
 
-### Server-Side Rendering Support (experimental)
```
andreas-karlsson, line 131: I think we can keep the "experimental" state as we're recommending the OpenFeature integration.

packages/react/README.md
```diff
@@ -180,76 +179,42 @@ export const ServerComponent = async () => {
 };
 ```
 
-#### Server and Client (experimental)
```
andreas-karlsson, line 183: I think we should keep this section as is. If we want a simplified React usage it should be an example of using the OpenFeature React SDK under "openfeature-web-provider" but not important to add in this PR.

andreas-karlsson (commented):
A great improvement of the docs over all, thanks!

README.md
```diff
@@ -4,98 +4,119 @@
...
+```
 
-You can set the context manually by using `setContext({})` or obtain a "child instance" of Confidence with a modified context by using `withContext({})`.
+### 3. Fetch a Flag
 
 ```ts
-confidence.setContext({ 'pants-color': 'yellow' });
-const childInstance = confidence.withContext({ 'pants-color': 'blue', 'pants-fit': 'slim' });
+const isEnabled = await client.getBooleanValue('feature.enabled', false, {
+  targetingKey: `your-targeting-key`,
```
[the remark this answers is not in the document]
nicklasl: What do you think about omitting the `targeting key` here and instead let this be:
```suggestion
  visitor_id: <unique id per visitor>
```
or 
```suggestion
  user_id: <user id for logged in users>
```

README.md
```diff
@@ -4,98 +4,121 @@
...
+The vanilla sdk can be used in cases where you want direct access to the Confidence SDK, including event tracking and custom context management.
+
+> **Learn more**: [SDK Documentation](./packages/sdk/README.md)
+
+## React Integration
+
+For React applications, use the dedicated React package that provides hooks and providers for seamless integration. This package is built on top of the direct SDK usage.
+
+> **Learn more**: [React Integration Documentation](./packages/react/README.md)
+
```
nicklasl, line 120: This is probably where we could mention the OF React SDK if we want to.

nicklasl (approved):
Looks good 👍

PR #296, opened 2026-03-06
test: add integration test for integer flag resolution

nicklasl:
## Summary
- Adds an integration test verifying the SDK correctly resolves an integer flag value from a resolve response with `intSchema`

## Test plan
- [x] New test passes locally


packages/sdk/src/Confidence.int.test.ts
```diff
@@ -99,6 +99,32 @@ describe('Confidence integration tests', () => {
     expect(await confidence.getFlag('flag2.str', 'goodbye')).toBe('hello again');
     expect(applyHandlerMock).not.toHaveBeenCalled();
   });
+  it('should resolve an integer value from a flag with intSchema', async () => {
```
nicklasl:
```suggestion
  it('should resolve an integer value from a flag with intSchema even if value contains delimiter', async () => {
```

PR #299, opened 2026-03-19
fix: align evaluation trace mappings with Swift and Java SDKs

fabriziodemaria:
## Summary

Fixes evaluation reason mappings in `evaluationTraceFromResult` to match the Swift and Java SDK implementations:

- **`FLAG_ARCHIVED` → `EVALUATION_REASON_DISABLED`** (was incorrectly falling through to `DEFAULT`)
- **`TARGETING_KEY_ERROR` → `EVALUATION_REASON_ERROR` + `EVALUATION_ERROR_CODE_TARGETING_KEY_MISSING`** (was falling through to `DEFAULT`)
- **`UNSPECIFIED` → `EVALUATION_REASON_UNSPECIFIED`** (default fallback was `DEFAULT` instead of `UNSPECIFIED`)
- Marks `millisecond_duration` on `Trace.traceData` as `[deprecated = true]` to match the backend proto

## Test plan

- [x] New test: `FLAG_ARCHIVED` produces `EVALUATION_REASON_DISABLED` trace
- [x] New test: `TARGETING_KEY_ERROR` produces `ERROR` + `TARGETING_KEY_MISSING` trace
- [x] New test: `NO_SEGMENT_MATCH` produces `EVALUATION_REASON_DEFAULT` trace
- [x] All 12 integration tests pass

Made with [Cursor](https://cursor.com)

### DIFF
--- a/packages/sdk/src/Confidence.ts
+++ b/packages/sdk/src/Confidence.ts
@@ -493,34 +493,33 @@ function evaluationTraceFromResult(evaluation: FlagEvaluation.Resolved<Value>): 
   const EvalReason = LibraryTraces_Trace_EvaluationTrace_EvaluationReason;
   const EvalError = LibraryTraces_Trace_EvaluationTrace_EvaluationErrorCode;
 
-  if (evaluation.reason === 'MATCH') {
-    return {
-      reason: EvalReason.EVALUATION_REASON_TARGETING_MATCH,
-      errorCode: EvalError.EVALUATION_ERROR_CODE_UNSPECIFIED,
-    };
+  switch (evaluation.reason) {
+    case 'MATCH':
+      return { reason: EvalReason.EVALUATION_REASON_TARGETING_MATCH, errorCode: EvalError.EVALUATION_ERROR_CODE_UNSPECIFIED };
+    case 'NO_SEGMENT_MATCH':
+    case 'NO_TREATMENT_MATCH':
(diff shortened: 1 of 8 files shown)


packages/sdk/src/Confidence.ts
```diff
@@ -493,34 +493,33 @@ function evaluationTraceFromResult(evaluation: FlagEvaluation.Resolved<Value>):
...
+  switch (evaluation.reason) {
+    case 'MATCH':
+      return { reason: EvalReason.EVALUATION_REASON_TARGETING_MATCH, errorCode: EvalError.EVALUATION_ERROR_CODE_UNSPECIFIED };
+    case 'NO_SEGMENT_MATCH':
+    case 'NO_TREATMENT_MATCH':
+      return { reason: EvalReason.EVALUATION_REASON_DEFAULT, errorCode: EvalError.EVALUATION_ERROR_CODE_UNSPECIFIED };
+    case 'FLAG_ARCHIVED':
+      return { reason: EvalReason.EVALUATION_REASON_DISABLED, errorCode: EvalError.EVALUATION_ERROR_CODE_UNSPECIFIED };
+    case 'TARGETING_KEY_ERROR':
+      return { reason: EvalReason.EVALUATION_REASON_ERROR, errorCode: EvalError.EVALUATION_ERROR_CODE_TARGETING_KEY_MISSING };
```
fabriziodemaria: There is a little discrepancy between "key missing" (OF) and "key error" (Confidence). But I think it's still close enough and more informative to use "EVALUATION_ERROR_CODE_TARGETING_KEY_MISSING" in this scenario

PR #302, opened 2026-03-24
fix: prevent apply request with empty flags array

nicklasl:
## Summary
- The applier's flush could fire even when no flags were added to pending, sending an empty flags array to the backend which returns HTTP 400
- Added early return guard in `flush()` to skip sending when there's nothing to apply

## Test plan
Verified with an instrumented test app using a custom fetch interceptor to trace all SDK network calls (resolve, apply, telemetry:upload). 
Before the fix: `FLAG_NOT_FOUND` evaluations triggered empty apply requests with `flags: []`. 
After: empty applies are eliminated and telemetry data is correctly flushed via the standalone `telemetry:upload` endpoint.

- [x] Added test: "should not send an apply request with empty flags when evaluating a nonexistent flag"
- [x] All existing tests pass

### DIFF
--- a/packages/sdk/src/FlagResolverClient.ts
+++ b/packages/sdk/src/FlagResolverClient.ts
@@ -110,6 +110,7 @@ export class FetchingFlagResolverClient implements FlagResolverClient {
   private readonly resolveTimeout: number;
   private readonly baseUrl: string;
   private readonly applyBaseUrl: string;
+  private readonly telemetry: Telemetry;
   private readonly traceConsumer: TraceConsumer;
   private readonly onEvaluation: EvaluationObserver | undefined;
   private readonly waitUntil: WaitUntil | undefined;
@@ -136,6 +137,7 @@ export class FetchingFlagResolverClient implements FlagResolverClient {
     cacheProvider,
     onEvaluation,
   }: FlagResolverClientOptions) {
+    this.telemetry = telemetry;
     this.traceConsumer = telemetry.registerLibraryTraces({
       library: LibraryTraces_Library.LIBRARY_CONFIDENCE,
       version: sdk.version,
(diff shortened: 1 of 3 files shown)


packages/sdk/src/FlagResolverClient.ts
```diff
@@ -228,12 +230,40 @@ export class FetchingFlagResolverClient implements FlagResolverClient {
...
+    }
+  }
+
+  private flushTelemetry(): void {
+    const monitoring = this.telemetry.getSnapshot();
+    if (monitoring.libraryTraces.length === 0) return;
+    const promise = this.uploadTelemetry(monitoring).catch(error => {
+      this.logger.info?.(`Confidence: Failed to upload telemetry: ${error.message}`);
+    });
+    // non-blocking: just keeps serverless runtimes alive until the upload completes
```
nicklasl:
```suggestion
```

packages/sdk/src/FlagResolverClient.ts
```diff
@@ -228,12 +230,39 @@ export class FetchingFlagResolverClient implements FlagResolverClient {
...
+    if (!resp.ok) {
+      throw new Error(`${resp.status}: ${resp.statusText}`);
+    }
+  }
+
+  private flushTelemetry(): void {
+    const monitoring = this.telemetry.getSnapshot();
+    if (monitoring.libraryTraces.length === 0) return;
+    const promise = this.uploadTelemetry(monitoring).catch(error => {
+      this.logger.info?.(`Confidence: Failed to upload telemetry: ${error.message}`);
```
fabriziodemaria, line 251: Warning level?

fabriziodemaria (approved):

packages/sdk/src/FlagResolverClient.ts
nicklasl: I debated that with myself but decided against it since it is just telemetry. nothing to alarm a customer about.

PR #354, opened 2026-06-10
fix(sdk): remove telemetry header and batch uploads

nicklasl:
## Summary
- Remove `X-CONFIDENCE-TELEMETRY` header from all requests — oversized headers were causing apply call failures (CORS/server rejection)
- Telemetry now flushes via the standalone `/v1/telemetry:upload` endpoint, triggered by buffer size (48KB, under the 64KB `keepalive` browser limit)
- Add `visibilitychange` flush with `keepalive: true` so telemetry isn't lost on page close
- Add `Confidence.close()` for server-side cleanup — flushes pending telemetry on shutdown. Called automatically by the OpenFeature server provider's `onClose` hook
- Early-return in the applier when no flags are pending, preventing empty flush cycles

## Context
Two features combined to create excessive network traffic:
1. Evaluation tracing (`6de4fda`) generates a trace on every `evaluate()` call
2. Standalone telemetry upload (`7780cc6`) fires on every empty applier tick

This caused a wall of `telemetry.upload` requests on every page, and the piggybacked telemetry header on apply requests grew unbounded — breaking apply calls.

## Test plan
- [x] All 159 unit/integration tests pass
- [x] Verified in React 19 example app: resolve + apply + telemetry upload only at buffer threshold (no spam)
- [x] Verified no `X-CONFIDENCE-TELEMETRY` header on resolve/apply requests
- [x] Verified `visibilitychange` triggers telemetry flush via simulated hidden state
- [x] Telemetry buffer size threshold tested with configurable `maxBufferBytes`
- [x] OpenFeature server provider `onClose` wired to `Confidence.close()`

### DIFF
--- a/packages/sdk/src/FlagResolverClient.ts
+++ b/packages/sdk/src/FlagResolverClient.ts
@@ -187,14 +187,16 @@ export class FetchingFlagResolverClient implements FlagResolverClient {
       this.cacheReadThrough = (_context, supplier) => supplier().then(response => ({ response, isFromCache: false }));
     }
 
-    const telemetryTimerId = setInterval(() => this.flushTelemetry(), 5_000);
-    if (typeof telemetryTimerId === 'object') telemetryTimerId.unref();
-    if (typeof document !== 'undefined') {
-      document.addEventListener('visibilitychange', () => {
-        if (document.visibilityState === 'hidden') {
-          this.flushTelemetry({ keepalive: true });
-        }
-      });
+    if (!telemetry.disabled) {
+      const telemetryTimerId = setInterval(() => this.flushTelemetry(), 5_000);
+      if (typeof telemetryTimerId === 'object') telemetryTimerId.unref();
+      if (typeof document !== 'undefined') {
+        document.addEventListener('visibilitychange', () => {
+          if (document.visibilityState === 'hidden') {
+            this.flushTelemetry({ keepalive: true });
+          }
+        });
+      }
     }
   }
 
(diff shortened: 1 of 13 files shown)


packages/sdk/src/FlagResolverClient.ts
```diff
@@ -259,10 +270,7 @@ export class FetchingFlagResolverClient implements FlagResolverClient {
     let [nextFlush, resolveNextFlush] = resolvablePromise();
 
     const flush = () => {
-      if (!pending.length) {
-        this.flushTelemetry();
-        return;
-      }
+      if (!pending.length) return;
```
nicklasl, line 273: revisit this.

andreas-karlsson: Yeah. It should be fine, cause I remember flushTelemetry is a noop if ther is none collected.

packages/sdk/src/Telemetry.ts
```diff
@@ -48,10 +56,12 @@ export class Telemetry {
...
-      traces.push({
-        id,
-        ...data,
-      });
+      const trace: LibraryTraces_Trace = { id, ...data };
+      traces.push(trace);
+      this.bufferBytes += LibraryTraces_Trace.encode(trace).finish().byteLength;
+      if (this.bufferBytes >= this.maxBufferBytes) {
+        this.onFlush?.();
+      }
```
andreas-karlsson: Should do this check _before_ adding the trace to traces no?

nicklasl: fixed

andreas-karlsson (approved):

PR #388, opened 2026-06-30
fix(sdk): use sendBeacon for visibilitychange flush

dwolmerud:
## Summary

The `visibilitychange` handler in `EventSenderEngine` uses `fetch({keepalive: true})` to flush queued events when the page becomes hidden. This is unreliable for page-unload delivery because:

- The 64KB keepalive budget is shared across all in-flight keepalive requests (including SDK telemetry)
- `fetch({keepalive})` has weaker browser guarantees than `navigator.sendBeacon()` for tab-close/navigate-away scenarios

We observed a ~95% drop in event samples (from 35-55k/day to ~2.5k/day) after switching from a transport that used `sendBeacon` to the Confidence SDK.

### Changes

Adds a `beaconFlush()` private method to `EventSenderEngine` that:
- Serializes the queue synchronously and calls `navigator.sendBeacon()` directly, bypassing the async `FetchBuilder` chain
- Caps the beacon batch at 25 events to stay within sendBeacon's ~64KB payload limit
- Falls back to `fetch({keepalive: true})` when `sendBeacon` is unavailable or the browser rejects the beacon
- No changes to the `upload()` public API

### Known limitation

`beaconFlush()` bypasses the configured `fetchImplementation` by design. sendBeacon must serialize and enqueue synchronously during page unload — the async FetchBuilder middleware chain cannot guarantee completion before the page is discarded. Consumers relying on `fetchImplementation` for proxying or instrumentation should be aware that visibilitychange flushes take a different path.

### Note

Session recordings (`csr`) use the same `keepalive` approach in their flush — this fix only addresses `EventSenderEngine` but the same pattern could be applied there.

- [x] All tests are passing

### DIFF
--- a/packages/sdk/src/EventSenderEngine.ts
+++ b/packages/sdk/src/EventSenderEngine.ts
@@ -81,7 +81,7 @@ export class EventSenderEngine {
     if (typeof document !== 'undefined') {
       document.addEventListener('visibilitychange', () => {
         if (document.visibilityState === 'hidden') {
-          this.flush({ keepalive: true });
+          this.beaconFlush();
         }
       });
     }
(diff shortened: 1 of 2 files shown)


nicklasl:
(e2e tests are expected to fail for external contributors/forks)

andreas-karlsson (approved):
LGTM!

nicklasl:
Ran the e2e tests locally against this branch — all 3 suites passed (24/24 tests green). :white_check_mark:

```
PASS  server-provider  packages/openfeature-server-provider/src/ConfidenceServerProvider.e2e.test.ts
PASS  web-provider     packages/openfeature-web-provider/src/ConfidenceWebProvider.e2e.test.ts
PASS  sdk              packages/sdk/src/Confidence.e2e.test.ts

Test Suites: 3 passed, 3 total
Tests:       24 passed, 24 total
```

PR #409, opened 2026-07-14
fix(csr): detect and report worker load failures caused by CSP

nicklasl:
## Summary

- Worker construction now catches both sync throws (Chrome `SecurityError`) and async `onerror` (Firefox) instead of silently swallowing the failure or hanging forever.
- Emits a `console.warn` naming the blocking CSP directive with actionable guidance.
- Adds a 10s welcome-handshake timeout so `createUploader` always resolves/rejects.

## Context

Customers with a `worker-src` CSP that blocks `data:` URLs get no recording and no signal — Chrome silently disables it, Firefox hangs indefinitely. This is Phase 1 of the CSP work; `blob:` fallback and `workerUrl` support follow separately.

## Test plan

- [x] Sync throw path (Worker + SharedWorker constructor blocked)
- [x] Async error path (Worker + SharedWorker `onerror` fires)
- [x] Welcome timeout (worker loads but never responds)
- [x] Custom `workerUrl` failure includes the URL in the warning
- [x] Full CSR test suite passes (119 tests)

### DIFF
--- a/csr/csr-common/src/uploader/create-uploader.ts
+++ b/csr/csr-common/src/uploader/create-uploader.ts
@@ -6,10 +6,12 @@ const STORAGE_TAB_ID = 'csr:tabId';
 const STORAGE_SESSION = 'csr:session';
 const STORAGE_COUNTER = 'csr:counter';
 const DEFAULT_SESSION_TTL_MS = 30 * 60 * 1000;
+const WELCOME_TIMEOUT_MS = 10_000;
 
 interface PortLike {
   postMessage(m: unknown): void;
   setHandler(cb: (data: unknown) => void): void;
+  onError(cb: (err: Error) => void): void;
 }
 
 interface WelcomeMessage {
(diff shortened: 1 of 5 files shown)


csr/csr-common/src/uploader/create-uploader.ts
```diff
@@ -230,25 +255,66 @@ async function openWorkerPort(
...
+
+    let worker: SharedWorker;
+    try {
+      worker = new SharedWorker(url, options);
+    } catch (err) {
+      throw workerLoadError(err, url);
+    }
+
+    worker.onerror = () => {
+      errorCb?.(workerLoadError(null, url));
```
andreas-karlsson: Why not pass the error event on to the workerLoadError?

csr/csr-common/src/uploader/create-uploader.ts
```diff
@@ -230,25 +255,66 @@ async function openWorkerPort(
...
+  const isDataUrl = url.startsWith('data:');
+  const hint = isDataUrl
+    ? 'This is likely caused by a Content Security Policy (CSP) that blocks `data:` in ' +
+      '`worker-src`. To fix this, either add `data:` to your `worker-src` directive, or ' +
+      'use the `workerUrl` option to serve the worker script from your own origin.'
+    : `Failed to load the worker script from ${url}. Check that the URL is reachable and ` +
+      'that your CSP `worker-src` directive allows it.';
+
+  // eslint-disable-next-line no-console
+  console.warn(`[@spotify-confidence/session-recording] Worker failed to load. ${hint}`);
```
andreas-karlsson: I'm hesitant on whether we always want to log this? Could we check and only log in dev-builds?

andreas-karlsson (dismissed):
Approved with two comments.

andreas-karlsson (approved):

PR #410, opened 2026-07-14
feat(csr): blob: URL fallback when data: worker is blocked by CSP

nicklasl:
## Summary

- When `new Worker(dataUrl)` / `new SharedWorker(dataUrl)` throws (CSP `worker-src` block), automatically retries with a `blob:` URL.
- Falls back to dedicated-worker mode since blob URLs are per-document (no cross-tab `SharedWorker` sharing).
- If both `data:` and `blob:` are blocked, emits a `console.warn` with actionable CSP guidance and throws.
- No fallback attempted when a custom `workerUrl` is provided — that's the customer's to debug.

## Context

Phase 2 of the CSP work. Most real-world CSPs already allow `worker-src blob:` (Sentry, PostHog, pdf.js all use blob workers), so this unblocks the majority of CSP-restricted customers with zero setup.

See also: #409 (Phase 1 — error detection/reporting)

## Test plan

- [x] Happy path: data: URL succeeds, no blob fallback
- [x] data: throws → blob: fallback used (dedicated worker)
- [x] SharedWorker data: throws → blob: dedicated fallback
- [x] Both data: and blob: blocked → throws with console.warn
- [x] Custom workerUrl → no fallback attempted
- [x] Full CSR test suite passes (118 tests)

### DIFF
--- a/csr/csr-common/src/uploader/create-uploader.test.ts
+++ b/csr/csr-common/src/uploader/create-uploader.test.ts
@@ -171,6 +171,44 @@ describe('createUploader', () => {
       expect(logs.some(l => l.includes('falling back to blob:'))).toBe(true);
     });
 
+    it('falls back to blob: when SharedWorker onerror fires asynchronously (Chrome CSP)', async () => {
+      let constructedUrl: string | undefined;
+      (globalThis as Record<string, unknown>).SharedWorker = class {
+        onerror: ((e: Event) => void) | null = null;
+        port = {
+          start() {},
+          onmessage: null as ((e: MessageEvent) => void) | null,
+          postMessage() {},
+        };
+        constructor() {
+          // Chrome doesn't throw — it fires onerror asynchronously
+          setTimeout(() => this.onerror?.(new Event('error')), 0);
+        }
+      };
+      (globalThis as Record<string, unknown>).Worker = class {
+        onerror: ((e: Event) => void) | null = null;
+        onmessage: ((e: MessageEvent) => void) | null = null;
+        postMessage() {}
+        constructor(url: string) {
+          constructedUrl = url;
+        }
+      };
+
+      const createUploader = await loadCreateUploader();
+      const logs: string[] = [];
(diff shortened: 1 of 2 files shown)


csr/csr-common/src/uploader/create-uploader.ts
```diff
@@ -211,35 +214,74 @@ function resolveMode(mode: 'shared' | 'dedicated' | 'auto'): 'shared' | 'dedicat
...
+      return { port: createDedicatedWorker(blobUrl), urlScheme: 'blob' };
+    } catch (blobErr) {
+      // Both data: and blob: blocked — nothing left to try.
+      // eslint-disable-next-line no-console
+      console.warn(
+        '[@spotify-confidence/session-recording] Worker failed to load. ' +
+          'Your Content Security Policy blocks both `data:` and `blob:` in `worker-src`. ' +
+          'To fix this, either add `blob:` to your `worker-src` directive, or ' +
+          'use the `workerUrl` option to serve the worker script from your own origin.',
+      );
```
andreas-karlsson: I think it'd be better if this was only logged in dev-mode.

andreas-karlsson (dismissed):

csr/csr-common/src/uploader/create-uploader.test.ts
```diff
@@ -60,11 +106,113 @@ describe('createUploader', () => {
...
+
+      const createUploader = await loadCreateUploader();
+      const logs: string[] = [];
+      createUploader({ ...DEFAULTS, workerMode: 'shared', debugLogger: m => logs.push(m) });
+      // Multiple ticks needed: hashSecret (microtask) → openWorkerPort's
+      // setTimeout probe (macrotask) → blob fallback creation
+      await tick();
+      await tick();
+      await tick();
+      await tick();
```
nicklasl: Ah! such an excellent solution. not at all the recipe for flaky tests!

andreas-karlsson (approved):

PR #415, opened 2026-07-17
feat(csr): log version drift between SDK and worker

nicklasl:
## Summary

Detects when a customer's self-hosted `confidence-worker.js` is out of sync with the installed SDK.

When customers use `workerUrl` (for strict CSP compliance), they copy the worker file to their static assets. If they later update the SDK without re-copying the file, the worker and SDK silently drift apart. This PR adds a content-hash handshake so the mismatch surfaces as an actionable `debugLogger` warning — recording still proceeds.

### How it works

1. `build-worker.mjs` computes a SHA-256 hash (16-char prefix) of the bundled worker code and writes it to `worker-hash.ts`
2. `emit-worker-file.mjs` prepends `globalThis.__WORKER_HASH__ = '<hash>'` to the standalone file
3. At runtime, the worker sends its hash in the welcome message; the tab compares against its own imported `WORKER_HASH`
4. Mismatch → `WORKER MISMATCH` warning via `debugLogger`, recording continues

### Why hashes instead of versions

csr-common and session-recording have independent version numbers. Using either package's version would show a number the customer doesn't recognize. A content hash sidesteps versioning entirely — if the worker code changed, the hash changed, regardless of which package bumped.

## Test plan

- [x] `workerHash` included in welcome when `globalThis.__WORKER_HASH__` is set
- [x] `workerHash` is undefined when not set (inlined worker path)
- [x] Mismatch warning logged when `urlScheme === 'custom'` and hashes differ
- [x] Warning for legacy workers without a hash (graceful upgrade nudge)
- [x] No warning when hashes match
- [x] No warning for data:/blob: URL schemes (inlined workers)
- [x] Build produces standalone file with hash prepended
- [x] Full CSR test suite passes (190 tests)


csr/csr-common/src/uploader/create-uploader.ts
```diff
@@ -154,6 +156,12 @@ export async function createUploader(opts: CreateUploaderOptions): Promise<Uploa
       ? `tab: welcome (${'sessionId' in welcome.result ? `sessionId=${welcome.result.sessionId}` : 'skipRecording'})`
       : `tab: dead reason=${welcome.reason}`,
   );
+  if (welcome.type === 'welcome' && urlScheme === 'custom' && welcome.workerHash !== WORKER_HASH) {
+    log?.(
+      'tab: WORKER MISMATCH — the self-hosted confidence-worker.js does not match the installed SDK. ' +
+        'Copy the updated file from node_modules/@spotify-confidence/session-recording/dist/confidence-worker.js',
+    );
+  }
```
sam-cook, line 164: Do we want to kill anything here? Or are we happy with just logging this and letting the two attempt to connect anyways?

sam-cook (approved):

csr/csr-common/src/uploader/create-uploader.ts
nicklasl: My take was that we should do a best effort in trying to record. It's better than failing completely IMO.

PR #419, opened 2026-07-27
feat(build): add CommonJS build output and package entrypoints (#156)

### LINKED ISSUE #156: Reconsider adding a CommonJS build output
It's still tricky to make Node execute TS with esm imports.

## Feature Suggestion

Add cjs output in the build.

vihaan-vijay:
## Summary
Fixes #156

This PR adds dual-module distribution support by adding CommonJS (CJS) package entrypoints and conditional exports alongside the existing ES Modules (ESM) build outputs across `@spotify-confidence/sdk` and related workspace packages.

## Problem
Projects relying on CommonJS `require()` syntax or TypeScript tooling running with default CommonJS settings (e.g. `ts-node`) encountered `ERR_PACKAGE_PATH_NOT_EXPORTED` errors because:
- The `package.json` manifest lacked top-level `"main"`, `"module"`, and `"types"` properties.
- The `"exports"` configuration map in `package.json` did not define a `"require"` conditional export pointing to CommonJS build artifacts.

## Changes Made
1. **Yarn Constraints & Package Manifests (`yarn.config.cjs` & `package.json`)**:
   - Updated `yarn.config.cjs` to enforce `"main"`, `"module"`, `"types"`, and `"exports"` invariants across workspace packages.
   - Configured `"main": "dist/index.cjs"`, `"module": "dist/index.mjs"`, and `"types": "dist/index.d.ts"`.
   - Added `"require": "./dist/index.cjs"` to the `"exports"` condition map.
2. **Cross-Platform Build Normalization (`rollup.base.mjs`)**:
   - Normalized Windows backslashes (`\`) to forward slashes (`/`) in `normalizePath` so Rollup bundling executes cleanly across Windows and POSIX environments.
3. **CJS Integration Test (`packages/sdk/src/cjs.test.ts`)**:
   - Added a unit test ensuring `@spotify-confidence/sdk` CommonJS artifacts can be required and cleanly expose exported interfaces (`Confidence`, `Value`, etc.).

## Verification
- **Unit Tests**: Ran `yarn test` — all 16 test suites (175 tests) passed cleanly.
- **Node.js CJS Resolution**: Verified `require('@spotify-confidence/sdk')` in Node.js runtime without errors.
- **Node.js ESM Resolution**: Verified `import { Confidence } from '@spotify-confidence/sdk'` in ESM runtime without errors.
- **Yarn Constraints**: Verified `yarn constraints` passes without constraint errors.

### DIFF
--- a/packages/openfeature-server-provider/package.json
+++ b/packages/openfeature-server-provider/package.json
@@ -44,8 +44,12 @@
   },
   "exports": {
     ".": {
-      "import": "./build/index.js",
-      "types": "./build/index.d.ts"
+      "import": "./dist/index.mjs",
+      "require": "./dist/index.cjs",
+      "types": "./dist/index.d.ts"
     }
-  }
+  },
+  "main": "dist/index.cjs",
+  "module": "dist/index.mjs",
+  "types": "dist/index.d.ts"
 }
--- a/packages/openfeature-web-provider/package.json
+++ b/packages/openfeature-web-provider/package.json
@@ -47,8 +47,12 @@
   },
   "exports": {
     ".": {
-      "import": "./build/index.js",
-      "types": "./build/index.d.ts"
+      "import": "./dist/index.mjs",
+      "require": "./dist/index.cjs",
+      "types": "./dist/index.d.ts"
     }
-  }
+  },
+  "main": "dist/index.cjs",
+  "module": "dist/index.mjs",
+  "types": "dist/index.d.ts"
 }
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -42,12 +42,14 @@
   },
   "exports": {
     ".": {
-      "import": "./build/index.js",
-      "types": "./build/index.d.ts"
+      "import": "./dist/index.mjs",
+      "require": "./dist/index.cjs",
+      "types": "./dist/index.d.ts"
     },
     "./server": {
-      "import": "./build/server.js",
-      "types": "./build/server.d.ts"
+      "import": "./dist/server.mjs",
+      "require": "./dist/server.cjs",
+      "types": "./dist/server.d.ts"
     }
   },
   "type": "module",
@@ -69,5 +71,8 @@
         "types": "./server.d.ts"
       }
     }
-  }
+  },
+  "main": "index.cjs",
+  "module": "index.mjs",
+  "types": "index.d.ts"
 }
--- a/packages/sdk/package.json
+++ b/packages/sdk/package.json
@@ -49,8 +49,12 @@
   "type": "module",
   "exports": {
     ".": {
-      "import": "./build/index.js",
-      "types": "./build/index.d.ts"
+      "import": "./dist/index.mjs",
+      "require": "./dist/index.cjs",
+      "types": "./dist/index.d.ts"
     }
-  }
+  },
+  "main": "dist/index.cjs",
+  "module": "dist/index.mjs",
+  "types": "dist/index.d.ts"
 }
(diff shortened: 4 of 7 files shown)


andreas-karlsson:
Thanks for digging into this, and sorry for the runaround — there's a wrinkle in how this repo publishes that makes the problem look different from the outside than it is.

**CJS output and the `require` export already ship.** The top-level `package.json` in this repo isn't what consumers see: `publishConfig` (set by `yarn.config.cjs`) is merged into the manifest at pack time, and it already carries `main`, `module`, `types`, and a `require` condition. Rollup has emitted both `.mjs` and `.cjs` since #241. Against the current release:

```
$ npm view @spotify-confidence/sdk@latest main exports
main: "dist/index.cjs"
exports: { ".": { types: "./dist/index.d.ts", import: "./dist/index.mjs", require: "./dist/index.cjs" } }

$ node -e "console.log(Object.keys(require('@spotify-confidence/sdk')))"
[ 'Value', 'Trackable', 'Closer', 'Confidence', 'pageViews', 'visitorIdentity', 'webVitals' ]
```

So plain `require()` works today, and the top-level fields this PR adds would be redundant with `publishConfig`.

**Why the top-level fields were removed in the first place.** The in-repo `exports` point at `./build/*.js` (raw `tsc` output) because that's what the monorepo itself resolves against — `dist/` only exists after `yarn bundle`, so repointing them means every local consumer needs a full rollup run instead of just `tsc`. And `main`/`module`/`types` were deliberately dropped from the top-level manifest in #241: they're single-entry legacy fields that can't express `@spotify-confidence/react`'s `./server` subpath, so they live in `publishConfig` where legacy resolvers need them. Relatedly, the `main: "index.cjs"` this PR adds to the react package points at a path that only exists between `prepack` (`cp dist/* .`) and `postpack` (`rm index.*`) — in the working tree it dangles.

One note on `packages/sdk/src/cjs.test.ts`: it requires `../dist/index.cjs`, so `yarn test` fails on a fresh clone — it only passes in CI because `Build.yml` runs `yarn bundle` first. The unit suite's resolver deliberately maps `@spotify-confidence/*` to `src/`; a check on built artifacts would need to be a separate pack/smoke step.

**That said, you were onto something real.** There *is* a remaining gap for TypeScript CJS consumers, which is probably what you or the reporter actually hit:

```
moduleResolution: node16   → error TS1479: the referenced file is an ECMAScript
                             module and cannot be imported with 'require'
moduleResolution: nodenext → clean (only because recent TS permits require(esm))
```

The package is `"type": "module"`, so `dist/index.d.ts` is read as ESM declarations, and there's no `dist/index.d.cts` for the `require` branch to resolve types from. Dropping a copied `index.d.cts` into `node_modules` makes `node16` compile clean, which confirms the cause. The fix is to emit `dist/index.d.cts` from the rollup dts step and use per-condition types (which also fixes `types` currently being listed last):

```json
"exports": {
  ".": {
    "import":  { "types": "./dist/index.d.ts",  "default": "./dist/index.mjs" },
    "require": { "types": "./dist/index.d.cts", "default": "./dist/index.cjs" }
  }
}
```

I'm going to close this one since the manifest changes would regress the local build — but the `normalizePath` backslash fix in `rollup.base.mjs` is a genuine bug worth having: `relative()` returns `\` on Windows, so the `path.startsWith('build/')` check in `createExternalTest` misfires. Rather than make you open a second PR for one line, I'll land it directly with you as co-author. And I'll track the `.d.cts` work separately so #156 gets properly closed for `ts-node`/`node16` users — that's the change that actually solves what you were chasing here.

andreas-karlsson:
Closing as described above — the manifest changes would regress local resolution, and `require()` already ships via `publishConfig`.

The `normalizePath` Windows fix is landing separately with you as co-author, and the real remaining gap is now tracked in #422. Thanks again for the investigation.

PR #426, opened 2026-08-04
feat: stateless remote client

andreas-karlsson:

Adds `ConfidenceClient` — a thin, stateless flag client for a remote resolver — as a new top-level export of `@spotify-confidence/sdk`. Ported from `confidence-resolver/openfeature-provider/js`, where it was originally written.

Flags only: resolve, evaluate, apply. No lifecycle, no background work, no cached state, so constructing one is free. Intended as the primitive the OpenFeature providers move onto once `Confidence` is retired — the deprecation notices are narrowed from "the SDK" to "the `Confidence` class" accordingly.

### New exports

- `ConfidenceClient` (+ `ConfidenceClient.Options` / `.ApplyResult`), `EvaluationContext`
- `FlagBundle` (+ `FlagBundle.evaluate`) — pure and dependency-free, so a server can resolve once and forward the bundle as JSON for the browser to evaluate

### Worth a look in review

- **Neither call rejects.** `resolve` returns an errored bundle that evaluates to defaults with an `ERROR` reason; `apply` returns `ApplyResult` rather than throwing, since the natural call site is fire-and-forget and an unhandled rejection kills the process on Node. `status` is exposed so callers can retry.
- **Signal only, no `timeout` option** — `AbortSignal.timeout()` covers deadlines and cancellation with one parameter. Deadline misses report `TIMEOUT`; a deliberate `abort()` does not. No built-in retry yet.
- **Context uses the wire spelling `targeting_key`**, not `targetingKey`.
- **Imports are deliberately limited** to generated protos, `./logger`, `./base64` and `./FlagBundle` — nothing from the `Confidence` graph, so the new code doesn't pin the old one alive.
- Adapted to this repo's older protos: `SDK_ID_JS_CONFIDENCE` (no thin-client id exists yet), and `MATERIALIZATION_NOT_SUPPORTED` / `UNRECOGNIZED_TARGETING_RULE` dropped from the reason union.

### Drive-by

- Extracted the base64 helpers to `base64.ts` and deleted the byte-identical private copy in `FlagResolution.ts`. Both relied on `globalThis.Buffer` resolving via `@types/node`, which only works because jest pulls it in — now typed locally (`TS7017` in editors otherwise).
- `api/sdk.api.md` regenerated; `release-please-config.json` gains `src/ConfidenceClient.ts` for its version literal.

- [x] All tests are passing

### DIFF
--- a/packages/sdk/README.md
+++ b/packages/sdk/README.md
@@ -3,10 +3,12 @@
 ![](https://img.shields.io/badge/lifecycle-beta-a0c3d2.svg)
 
 > [!NOTE]
-> This Confidence standalone SDK is being phased out. For new integrations, we recommend using the OpenFeature APIs directly:
+> The standalone `Confidence` class is being phased out. For new integrations, we recommend using the OpenFeature APIs directly:
 >
 > - **Client-based (SPA)**: Use [@spotify-confidence/openfeature-web-provider](https://github.com/spotify/confidence-sdk-js/blob/main/packages/openfeature-web-provider/README.md)
 > - **Server**: Use [@spotify-confidence/openfeature-server-provider-local](https://github.com/spotify/confidence-resolver/tree/main/openfeature-provider/js/README.md), which resolves flags in-process with close to zero latency
+>
+> [`ConfidenceClient`](#confidenceclient) is a low-level client for a remote resolver, and will become the engine behind the providers above. Most integrations should use a provider rather than calling it directly.
 
 JavaScript implementation of the Confidence SDK, enables event tracking and feature flagging capabilities in conjunction with the OpenFeature Web SDK.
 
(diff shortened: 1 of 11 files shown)


vahidlazio (approved):

fabriziodemaria (commented):
Product-level review — focusing on packaging, naming, and documentation completeness.

packages/sdk/src/index.ts
```diff
@@ -9,3 +9,7 @@ export * from './Closer';
 export * from './types';
 export { SimpleFetch } from './fetch-util';
 export { CacheOptions, CacheScope } from './flag-cache';
+// Named explicitly rather than `export *`, so this surface stays deliberate.
+// ConfidenceClient carries its Options and ApplyResult types in a merged namespace.
+export { ConfidenceClient, EvaluationContext } from './ConfidenceClient';
```
fabriziodemaria, line 14: **Thin client inside a large SDK package**

`ConfidenceClient` is described as a zero-dependency, stateless primitive — "no lifecycle, no background work, no cached state." But it ships inside `@spotify-confidence/sdk`, which carries event tracking, context management, caching, visitor identity, page views, web vitals, and all their dependencies.

A worker that only needs `resolve` + `evaluate` pulls in the entire SDK. The `FlagBundle` module header even says it "has no imports on purpose: it is reached for from code that must not depend on the rest of the package" — but consumers can't benefit from that isolation because tree-shaking isn't reliable across all bundlers and the install still brings everything.

Should this be `@spotify-confidence/client` (or similar) as a standalone package, with `@spotify-confidence/sdk` depending on it? That would match the stated goal of being "the primitive the providers are moving onto."

packages/sdk/src/ConfidenceClient.ts
```diff
@@ -0,0 +1,266 @@
...
+ * Confidence resolver Worker reached via service binding, or
+ * `resolver.confidence.dev` over HTTP.
+ *
+ * There is no lifecycle, no background work and no cached state: constructing
+ * one is free, so it can be created per request or shared at module level —
+ * it makes no difference.
+ * @public
+ */
+// eslint-disable-next-line @typescript-eslint/no-redeclare
+export class ConfidenceClient {
```
fabriziodemaria, line 85: **Naming: `ConfidenceClient` vs `Confidence`**

Two classes in the same package with very similar names but completely different APIs and mental models. `Confidence` has `setContext`, `getFlag`, `evaluateFlag`, `track`, `close`. `ConfidenceClient` has `resolve` and `apply` — a different level of abstraction.

For someone scanning the exports or autocomplete, the relationship isn't obvious. "Client" doesn't convey "stateless remote resolver." Something like `ConfidenceResolver` or `RemoteResolver` would signal the role more clearly.

Not a blocker — but once this name is in the public API it's hard to change, so worth a moment to consider.

packages/sdk/src/FlagBundle.ts
```diff
@@ -0,0 +1,183 @@
...
+   * A pure function with no I/O — it works on a bundle that was JSON-forwarded
+   * from the server, so a browser can evaluate without resolving again. Never
+   * throws: errors surface as the default value with an `ERROR` reason.
+   *
+   * @param bundle - a bundle from `ConfidenceClient.resolve`
+   * @param flagKey - `'my-flag'` or a dot path into the value, `'my-flag.some.field'`
+   * @param defaultValue - returned whenever the flag cannot be evaluated
+   * @param logger - optional logger for evaluation failures
+   */
+  export function evaluate<T extends Value>(
```
fabriziodemaria, line 84: **`FlagBundle` as a namespace-with-evaluate is an unusual API shape**

`FlagBundle` is both an interface (the data) and a namespace (with `evaluate` as a static function):

```ts
FlagBundle.evaluate(bundle, 'my-flag.title', 'default')
```

The reason is sound (keeping the bundle as plain JSON so it survives serialization), but this pattern is TypeScript-specific and won't map cleanly if the API is ever mirrored in other languages. A standalone `evaluate(bundle, ...)` export or a wrapper class with a method would be more conventional.

Worth confirming this is the shape you want to commit to publicly.

packages/sdk/src/ConfidenceClient.ts
```diff
@@ -0,0 +1,266 @@
...
+};
+
+/**
+ * Evaluation context, passed through to targeting verbatim.
+ *
+ * Note the wire spelling `targeting_key` — this is the resolver's contract, not
+ * OpenFeature's `targetingKey`.
+ * @public
+ */
+export type EvaluationContext = {
```
fabriziodemaria, line 28: **`EvaluationContext` introduces a second context type with subtly different casing**

The package already exports `Context` (from the `Confidence` class). Now it also exports `EvaluationContext`, which uses `targeting_key` (snake_case, matching the wire format) while `Context` / OpenFeature use `targetingKey` (camelCase).

The README calls this out, but having two context types with different key casing in the same package is a footgun — especially for consumers migrating from `Confidence` to `ConfidenceClient`, or working with both.

packages/sdk/README.md
```diff
@@ -20,6 +22,145 @@ To add the packages to your dependencies run:
...
+const bundle = await client.resolve([], { targeting_key: userId }, { apply: false });
+return { props: { bundle } }; // serialize into the page
+
+// --- browser ---
+import { FlagBundle } from '@spotify-confidence/sdk';
+
+const showBanner = FlagBundle.evaluate(bundle, 'promo-banner.enabled', false);
+if (showBanner.shouldApply) {
+  await fetch('/api/apply', { method: 'POST', body: JSON.stringify({ flag: 'promo-banner' }) });
+}
```
fabriziodemaria, line 69: **The apply-from-browser story is incomplete**

This example shows the browser calling `fetch('/api/apply', ...)`, but the server-side handler for that endpoint is left as an exercise. The browser can't call `client.apply()` directly (no client secret), so the consumer needs to build their own proxy endpoint that receives the token + flag name and calls `client.apply(resolveToken, flagNames)`.

Consider either showing both sides of the `/api/apply` route, or adding a note that the browser-side apply requires a server proxy that calls `client.apply`.

andreas-karlsson: The client can call apply exposing it's secret (that's what Confidence) does, but we don't recommend it any longer.

PR #451, opened 2026-09-08
feat(csr): retire force recording semantics

umberto-sonnino:
Retires the Session Recording SDK's public force-recording behavior while preserving when manual recording starts.

### Before → after

```diff
 initSessionRecorder
   automatic mode
     -> initAndRecord()

   manual mode
     -> wait for SessionRecorder.start()
-    -> initAndRecord(forceRecord: true)
+    -> initAndRecord()

 createUploader -> worker hello -> CsrClient.initSession()
-  POST { clientSecret, context, forceRecord: true }
+  POST { clientSecret, context }

 backend response
   credentials
     -> start recording

   skipRecording
     -> remain in the non-recording state

 worker already in the non-recording state
-  forceRecord hello -> send another InitSession request
+  later hello -> return the existing skip result
```

This keeps manual mode useful for consent or feature-flag timing, but starting manually no longer bypasses backend eligibility, targeting, allocation, sampling, or admission.

Automatic initialization, context, credentials, `skipRecording`, and error handling are unchanged. The SDK does not retry after a skip, HTTP 429, backend failure, or overload response.

### Compatibility

This intentionally removes `forceRecord` from the low-level `@spotify-confidence/csr-common/uploader` API. The public `SessionRecorder` shape remains unchanged.

The backend accepts and ignores the deprecated request field, so previously published SDK versions remain wire-compatible while new SDK requests stop sending it.

### Supporting changes

- Adds a compile-time assertion that `CreateUploaderOptions` does not expose `forceRecord`.
- Covers automatic and manual initialization behavior.
- Covers safe handling of a backend skip.
- Verifies that skips and HTTP 429/500/503 responses do not create another `InitSession` request.
- Updates the manual-mode documentation.
- Regenerates the worker bundle from its TypeScript sources.

The regenerated worker also synchronizes the existing session-token debug-log redaction from source into the generated output.

### Validation

- `yarn bundle`
- `yarn test` — 220 tests passed
- `yarn verify:types`
- `yarn build:csr`
- `yarn test:csr` — 173 tests passed
- `yarn lint`
- `yarn lint:csr`
- `yarn format:check`
- Compile-time `CreateUploaderOptions` compatibility check
- `git diff --check`

- [x] All tests are passing
- [ ] Tested in a corresponding example app — not run; behavior is covered at the public recorder, uploader, request, and worker boundaries


csr/session-recording/src/index.ts
```diff
@@ -57,7 +57,7 @@ export interface InitSessionRecorderOptions {
   context?: ClientContext;
   /**
    * `'automatic'` (default) — starts recording as soon as the session is established.
-   * `'manual'` — does nothing until `start()` is called, bypassing sampling and targeting rules.
+   * `'manual'` — does nothing until `start()` is called, then uses normal backend eligibility.
```
nicklasl, line 60: I'm a bit concerned by what the `start()` function actually means if there is no guarantee of a recording being started.

I’d rename `start()` to **`requestRecording()`** and return a promise with an explicit outcome. The caller requests; the backend grants admission.

  ```ts
  type RecordingStartResult =
    | { status: 'recording' }
    | { status: 'skipped' }
    | { status: 'failed'; reason: string }
    | { status: 'stopped' };

  interface SessionRecorder {
    requestRecording(): Promise<RecordingStartResult>;
    stop(): void;
    readonly isRecording: boolean;
    // tag() and measure() unchanged
  }
  ```

  ```ts
  const result = await recorder.requestRecording();

  if (result.status === 'skipped') {
    // Backend declined recording.
  }
  ```

  The behavior would be consistent across both modes:

  - **Manual:** the first call initiates the request.
  - **Automatic:** initialization initiates it; calling `requestRecording()` lets you await its outcome.
  - **Repeated calls:** share the existing attempt, without retrying skipped or failed requests.
  - **Stopped:** calls resolve to `stopped`; stopping during initialization also resolves the pending attempt to `stopped`.

  The result describes the attempt’s outcome; `isRecording` describes whether recording is active **now**. A later disconnect doesn’t change a previously returned result.

  This also removes the current “no-op in automatic mode” special case. I’d keep skip reasons out until the backend actually provides them, and defer state subscriptions until there’s a concrete need.

umberto-sonnino: I think you're absolutely right on this, and we should update the API to this design. But, shouldn't that come as a separate piece of work?

This PR is just for removing the flag that's effectively not doing anything anymore, so it's just for cleaning that up rather than updating the contract. I

nicklasl: I talked to @andreas-karlsson and he thought we should keep the public API as is so we can disregard this comment.

nicklasl (dismissed):
Let's ship this. 
I suggest we do not mark it as a breaking change. The breakage is not in the "customer facing package" (session-recording). csr-common is only public because of convenience.

nicklasl (approved):

PR #452, opened 2026-09-09
fix: strip query and hash from recording URLs

nicklasl:

rrweb Meta events and the session referrer could retain query parameters and fragments containing private data. Strip both before recording while preserving route parameterization, including for relative or malformed URLs.

Validation: all 184 CSR tests pass; csr-common and csr-recorder typechecks, changed-file lint, and formatting checks pass.

- [x] All tests are passing

### DIFF
--- a/csr/csr-common/src/url.ts
+++ b/csr/csr-common/src/url.ts
@@ -1,3 +1,8 @@
+/** Remove query strings and fragments, including from relative or malformed URLs. */
+export function stripUrlQueryAndHash(url: string): string {
+  return url.split(/[?#]/, 1)[0];
+}
+
 /**
  * Extract only the pathname from a URL, stripping origin, query string, and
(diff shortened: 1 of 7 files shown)


csr/csr-common/src/url.ts
```diff
@@ -1,3 +1,8 @@
+/** Remove query strings and fragments, including from relative or malformed URLs. */
+export function stripUrlQueryAndHash(url: string): string {
+  return url.split(/[?#]/, 1)[0];
+}
```
sam-cook, line 4: Could we use `new Url()` for this, and remove the query?

sam-cook: Nevermind, this actually handles malformed URLs a bit better.

sam-cook (approved):

PR #455, opened 2026-09-09
feat(csr): authenticate recording WebSockets with subprotocols

sam-cook:

Recording WebSocket connections currently include the session token in the URL. Send it through `Sec-WebSocket-Protocol` as `recording.v1` plus `auth.<sessionToken>` instead, keeping the credential out of the URL. The issued token is already safe for this header, so it is sent unchanged. Reject empty tokens, unsupported characters, and tokens longer than 4096 characters.

- [x] All tests are passing

The local recording dev server was tested as described above; the repository example app was not run.


AHB (dismissed):

csr/csr-common/src/uploader/worker/core.test.ts
```diff
@@ -222,9 +227,96 @@ describe('worker/core', () => {
...
+      const logs = port.received.filter(isType('log')).map(message => (message as { msg: string }).msg);
+      expect(logs.length).toBeGreaterThan(0);
+      for (const log of logs) {
+        expect(log).not.toContain(token);
+      }
+    });
   });
 
   describe('lifecycle after active', () => {
+    it('uses a fresh header credential after a stale session hint fails', async () => {
```
umberto-sonnino: Not sure how big a deal this is, but this test models a pre-open protocol failure, but on https://ghe.spotify.net/konfidens/epx-session-replay/pull/347 we return `101` with `recording.v1` and then close `4401` for a valid token whose session is missing or closed. 

Here `ready()` has already resolved, so core adopts the cached hint and the later `4401` is terminal; it never calls `InitSession`. Should we add coverage for this `recording.v1` >> followed by an immediate `4401`?

umberto-sonnino (dismissed):

umberto-sonnino (approved):

AHB (approved):

PR #456, opened 2026-09-10
fix: revive expired recording sessions on activity

AHB:

When the backend ends a session (for example after an inactivity timeout), the SDK
previously treated the closed transport as terminal: the worker moved to `dead`,
broadcast that to every tab, and `onTerminate` dismantled the recorder. Nothing was
captured for the rest of the page's life.

The practical effect was total, silent recording loss — sit idle long enough for the
session to be closed, come back, and no further activity was ever recorded until a
full page reload.

### What changed

- **`dead` is no longer the destination for a closed transport.** Two new worker phases,
  `interrupted` and `restarting`, replace the terminal transition.
- **Revival is driven by user activity, not a timer.** A frame only triggers recovery when
  `userActivity === true`, so a backgrounded or abandoned tab never re-establishes a
  session on its own.
- **`restartSession`** establishes a fresh session, reopens the transport, and notifies
  every attached port with `session-restarted`, carrying `adoptedFromSessionId` so the new
  session records which one it succeeded.
- **Capture restarts with a fresh full snapshot.** The tab resets its frame counter and
  calls the new `onSessionRestart` hook, which tears down and restarts the recorder.
  Without a new snapshot the replacement recording would be incremental mutations with
  nothing to apply them to, and therefore unreplayable.
- **Recovery retries are bounded.** A failed restart backs off, doubling per failure up to
  a 5 minute ceiling. Without a ceiling a persistently failing backend would draw one
  `initSession` per 5s of active use, indefinitely, on every site embedding the SDK.
  Transient failures still recover on the next interaction.
- **Every hello gets answered.** Ports track whether their hello has been answered, and a
  failed recovery answers the ones still waiting so they fail fast. See below.
- **Idle-resume failures recover too.** `resumeTransport` previously died on a stale token
  and now falls through to a restart.

### The unanswered-hello case

Worth calling out because it is the sharp edge of adding a non-terminal phase, and it cost
a round of review to find.

A tab whose hello landed while the worker sat in `interrupted` triggered a restart, and
when that restart failed, the flush had no branch for the phase it returned to — so the
hello went unanswered. That tab hung until its welcome timeout, `createUploader` threw, and
recording was disabled for the rest of the page's life. It could not recover later either:
with no uploader it could never emit the activity frame that drives recovery.

On a multi-page site, where every navigation is a fresh hello, that turned a backend
incident into lasting per-page data loss rather than a gap. Before `interrupted` existed
the same tab was told `dead` immediately and failed fast, so this was a regression rather
than a pre-existing gap.

The `answered` flag is what makes the fix safe: the flush walks every port, and a late
`dead` sent to an already-established tab dismantles a healthy recorder.

### Drive-by fix

`onClose` and `onStateChange` now verify the callback belongs to the current transport
(`state.transport !== transport`) before mutating state, so a superseded transport's
callbacks can no longer clobber a newer one.

### Testing

- **`WebSocketTransport`**: the backend closes a session with 1001 (`ENDPOINT_UNAVAILABLE`),
  which is a graceful drain rather than a terminal close — so the transport reconnects, and
  only the refused reconnect (whose token now names a closed session) ends it. Covered at
  this layer because it is a transport concern; the `reconnect-failed` reason is what proves
  a reconnect was attempted rather than the drain itself being fatal.
- **`worker/core`**: a closed transport followed by a passive frame does nothing; a
  subsequent active frame establishes a new session, emits `session-restarted` with the
  previous session id, and resumes streaming — with no `dead` message emitted.
- **`worker/core`**: `session-restarted` reaches every attached port with identical
  contents.
- **`worker/core`**: a hello arriving while recovery is failing is answered with a `dead`,
  and the already-established tab does not receive one.
- **`worker/core`**: a failed restart does not spend another `initSession` on the following
  active frames while inside the backoff window.
- **`createUploader`**: a `session-restarted` message resets the frame counter to 0 and
  stamps the next frame with `adoptedFromSessionId`.
- **`initSessionRecorder`**: `onSessionRestart` stops the old recorder and starts a new one
  with identical config, leaving `isRecording` true.

Both new negative tests were confirmed to fail when their fix is reverted, rather than
only passing alongside it.

`worker-script.ts` regenerated, since `core.ts` is bundled into it.

### Manual verification

Exercised against production with a purpose-built harness bundled from this branch (not a
published version), using a 100%/100% recording policy:

- session established, closed via `sessions:closeSession` — which reaches
  `closeAndPublish` -> `closeSessionAndConnections` -> `CloseWebSocketFrame(ENDPOINT_UNAVAILABLE)`,
  the same path the idle janitor uses, so it reproduces a real idle timeout in seconds;
- recovery on the next user interaction produced a **new** sessionId;
- both recordings replay from the first frame, confirming `onSessionRestart` emits a fresh
  full snapshot. The unit test can only assert `record()` was called again with the same
  config, so this is the part that needed a real browser.

The observed close reason was `reconnect-failed code=4401`, confirming the two-hop path:
the backend's 1001 is treated as a graceful drain, the transport reconnects, and the
reconnect is refused because the token now names a closed session. Worth knowing that a
session close therefore costs one extra WebSocket handshake.

### Known behaviour change

The backoff bounds *failed* restarts, not successful ones. A connection that drops and
recovers repeatedly now yields a new session — and so a new Recording, a fresh sampling
decision, and an analysis job — each time, where a terminal close previously capped it at
one session per page load. That is inherent to recovering the session rather than a defect,
but it is a volume change worth expecting.

- [x] All tests are passing
- [x] Verified manually against production (see below)


csr/csr-common/src/uploader/worker/core.ts
```diff
@@ -303,23 +327,92 @@ async function resumeTransport(client: Client, sessionId: string, sessionToken:
...
+  }
+
+  wireTransport(transport);
+  state = {
+    phase: 'active',
+    client,
+    transport,
+    sessionId: result.sessionId,
+    sessionToken: result.sessionToken,
+  };
```
nicklasl, line 376: Could a frame from the previous session arrive after we switch to active but before its tab has processed session-started? 
I guess we'll need some way to catch those frames, or flush somehow. Otherwise they might end up in the new recording with messed up counters and dom references (even before the dom snapshot).

csr/session-recording/src/index.ts
```diff
@@ -150,6 +150,11 @@ export function initSessionRecorder(options: InitSessionRecorderOptions): Sessio
           stopRecorder = null;
           stopped = true;
         },
+        onSessionRestart: () => {
+          if (stopped || sendEvent === null) return;
+          stopRecorder?.();
+          stopRecorder = record(sendEvent, recordingConfig);
+        },
```
nicklasl, line 174: Will we get the existing flag evaluations in this new recording? I am not sure if `observeFlags()` will emit the flag evaluations for this case which mean we might miss any existing flag evaluations that affect the experience.

csr/csr-common/src/uploader/worker/core.ts
```diff
@@ -192,20 +210,25 @@ function onHello(handle: PortHandle): void {
       void resumeTransport(client, sessionId, sessionToken).then(flushPendingWelcomes);
       return;
     }
+    case 'interrupted': {
+      const { client, previousSessionId, nextRetryDelayMs } = state;
+      state = { phase: 'restarting', client, previousSessionId };
+      void restartSession(client, previousSessionId, nextRetryDelayMs).then(flushPendingWelcomes);
```
nicklasl, line 235: we are not respecting the retryAt here, or? I think that could be a reasonable thing (to ignore the retry backoff), when the user has made a navigation for example.

csr/csr-common/src/uploader/types.ts
```diff
@@ -53,6 +53,8 @@ export interface CreateUploaderOptions {
   }) => void;
   /** Called once when recording is permanently dead. SDK should dismantle the recorder. */
   onTerminate?: (info: { reason: string }) => void;
+  /** Called after an expired backend session is replaced so capture can emit a fresh full snapshot. */
+  onSessionRestart?: () => void;
```
nicklasl: is there a valid reason to keep this optional?

AHB: Nah, just that some of the other ones were optional, but seeing as we are the only people using it and it is implemented then can be required np

csr/session-recording/src/index.ts
AHB: Nope! We were not, it would have only got new ones (which isn't that helpful). Made it replay them now

csr/csr-common/src/uploader/worker/core.ts
AHB: Yes it could have snuck it! `awaitingReset` is now on the port, and it will drop anything that isn't the first one of the new session

csr/csr-common/src/uploader/worker/core.ts
AHB: Yes in this case the retryAt is being skipped. A frame being received does respect that, because that is a lot noisier and potentially more frequent.

for the `onHello` though, think it make sense to skip that, considering it is a user navigation and new page. If we did respec tit, we'd probably timeout on the hello and kill the recording for the entire tab

nicklasl (approved):

PR #463, opened 2026-09-14
fix(csr-recorder): ignore blocked iframe attachments

sam-cook:
## Summary

- Prevent `csr-recorder` from emitting iframe document attachments for blocked iframe placeholders.
- Keep normal iframe attachments unchanged.
- Remove tracked iframe IDs with their deleted DOM subtree, so the tracking data stays bounded.

## What was wrong

An iframe is a page inside a page. We intentionally replace a blocked iframe with an inert `div`, but rrweb could still emit a later event that treated the same node as an iframe. The browser then tried to put a document inside a `div`, rejected the DOM update, and the replay became white from its first frame.

```text
Before
blocked iframe #675 -> recorder changes it to div #675
                     -> rrweb attaches a document to #675
                     -> browser rejects document-inside-div
                     -> replay DOM is empty

After
blocked iframe #675 -> recorder changes it to div #675
                     -> recorder remembers #675 is a placeholder
                     -> recorder drops only its document attachment
                     -> replay DOM continues normally
```

This is a forward-only fix. It prevents new recordings from containing the invalid event pair. It does not repair existing recordings.

- [x] All tests are passing

Generated with [Codex](https://openai.com/codex/).


AHB (approved):

sam-cook:
Browser A/B verification completed with real Google Chrome and the current `csr-player`.

The fixture contained one blocked iframe, one normal iframe, and visible page content. Both iframes loaded after recording started. I ran each revision five times.

| Revision | Runs | Attachments to blocked placeholder | Attachments to normal iframe | Page marker visible in replay |
| --- | ---: | ---: | ---: | --- |
| `main` at `b1088ba` | 5/5 | 2 | 2 | Yes |
| PR at `961ef51` | 5/5 | 0 | 2 | Yes |

This verifies that the PR removes the invalid event pair and does not remove normal iframe attachments. The result was stable across all runs.

The small fixture did not reproduce the full white replay on `main`, so this test does not independently reproduce that visual symptom. The two production recordings remain the evidence that connects this invalid event pair to the empty replay DOM.

